PR #18, opened 2020-04-23
Remove Type Storage

MasonProtter:
Closes https://github.com/shashi/SymbolicUtils.jl/issues/17

In the process of writing this, I noticed some problems with the way `flatten_term` and `sort_args` were implemented (always convert ed output symtype to `Number`) so I took the liberty of fixing that as well. 

Can be factored into a separate PR if desired.

### DIFF
--- a/src/simplify.jl
+++ b/src/simplify.jl
@@ -175,24 +175,37 @@ pow(x::Symbolic,y) = y==0 ? 1 : Base.:^(x,y)
 function flatten_term(⋆, args)
     # flatten nested ⋆
     flattened_args = []
+    types = Type[]
     for t in args
         if t isa Term && operation(t) === (⋆)
             append!(flattened_args, arguments(t))
         else
             push!(flattened_args, t)
         end
+        push!(types, symtype(t))
     end
-    Term(⋆, Number, flattened_args)
+    T = if (⋆) ∈ (+, *)
+         promote_type(types...)
+    else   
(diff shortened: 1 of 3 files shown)


MasonProtter:
Looks like CI isn't running, but tests pass locally.

src/simplify.jl
```diff
@@ -175,24 +175,37 @@ pow(x::Symbolic,y) = y==0 ? 1 : Base.:^(x,y)
...
             push!(flattened_args, t)
         end
+        push!(types, symtype(t))
     end
-    Term(⋆, Number, flattened_args)
+    T = if (⋆) ∈ (+, *)
+         promote_type(types...)
+    else
+        # This will just give Any for abstract types. We need to roll our own or use concrete types concrete types
+        Base.promote_op(f, types...)
```
shashi, line 191: `promote_symtype` already exists right?

shashi: We should just probably call this by default in a `Term(f, args)` constructor. Then we can just call `Term` here.

I think I used `Number` here just as a stop-gap, and it's really fast :)

MasonProtter: Oh, I see yeah I missed  `promote_symtype`. Should we add a fallback to `promote_symtype` which does `promote_op(f, symtype.(args)...)`?

> We should just probably call this by default in a Term(f, args) constructor. Then we can just call Term here.

Sounds good.

MasonProtter:
Tests pass with new change.

PR #20, opened 2020-04-24
WIP Interface for `simplify`

shashi:
This should allow MTK and others to call `simplify` by defining 3 tiny accessor methods.

This is what @MasonProtter and I agreed on in a discussion today. To summarize:

1. This seems to be the shortest path to getting MTK to work with this right away even forgetting all the idiosyncratic differences. (the conversion should be cheap, we construct the whole tree several times during a `simplify` so this is no big deal).
2. work on the tracing analysis in can be done on this in a separate package without Symbolic <: Numeric decisions
3. Mike Innes wanted to use the rules system on Mjolnir IR, which should become possible with this.

### DIFF
--- a/src/simplify.jl
+++ b/src/simplify.jl
@@ -78,7 +78,7 @@ end
 <ₑ(a::T, b::S) where {T, S} = T===S ? isless(a, b) : nameof(T) < nameof(S)
 
 function <ₑ(a::Term, b::Term)
-    if arglength(a) <= 2 && arglength(b) <= 2
+    if 0 < arglength(a) <= 2 && 0 < arglength(b) <= 2
         # e.g. a < sin(a) < b ^ 2 < b
         @goto compare_args
     end
--- a/src/simplify.jl
+++ b/src/simplify.jl
@@ -166,17 +166,6 @@ end
 
 ### Simplification rules
 
-to_symbolic(x::Symbolic) = x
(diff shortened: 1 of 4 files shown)


src/simplify.jl
```diff
@@ -166,7 +166,18 @@ end
...
-simplify(x, rules=SIMPLIFY_RULES) = RuleSet(rules)(x)
+to_symbolic(x::Symbolic) = x
+function to_symbolic(x)
+    if !istree(x)
+        return x
+    end
+
+    Term(operation(x),
+         symtype(x),
+         map(to_symbolic, args(x)))
```
MasonProtter, line 177: This should be `Term{symtype(x)}(operation(x), map(to_symbolic, args(x)))` right?

shashi: Well you're right! I'm about to test this :-p

src/symbolic.jl
```diff
@@ -67,10 +67,18 @@ function to_symbolic(x)
         return x
     end
 
+    op = operation(x)
+
+    # This is a valid transformation, it helps ModelingToolkit IR
+    # get simpler for many constant related rules to apply
+    if op === identity && length(arguments(x)) === 1
+        return to_symbolic(arguments(x))
+    end
```
shashi, line 76: Maybe we can add this as a rule!

PR #24, opened 2020-04-25
Remove multiple_of

YingboMa:
Fix https://github.com/shashi/SymbolicUtils.jl/issues/23

### DIFF
--- a/src/rulesets.jl
+++ b/src/rulesets.jl
@@ -36,15 +36,17 @@ BASIC_NUMBER_RULES = let
 end
 
 TRIG_RULES = let
-    [[[@rule trig_f(~~x + ~y::multiple_of(2π) + ~~z) => trig_f(+(~~x..., ~~z...))
-       
-       @rule trig_f(~~x + ~y::multiple_of(2π) * ~n::oftype(Integer) + ~~z) => trig_f(+(~~x..., ~~z...))
-       @rule trig_f(~~x + ~n::oftype(Integer) * ~y::multiple_of(2π) + ~~z) => trig_f(+(~~x..., ~~z...)) # make multiplication commute
+    [
+     # https://github.com/shashi/SymbolicUtils.jl/issues/23
+     #[[@rule trig_f(~~x + ~y::multiple_of(2π) + ~~z) => trig_f(+(~~x..., ~~z...))
+     #
+     #  @rule trig_f(~~x + ~y::multiple_of(2π) * ~n::oftype(Integer) + ~~z) => trig_f(+(~~x..., ~~z...))
+     #  @rule trig_f(~~x + ~n::oftype(Integer) * ~y::multiple_of(2π) + ~~z) => trig_f(+(~~x..., ~~z...)) # make multiplication commute
 
-       @rule trig_f(~~x + ~y::multiple_of(2π) * ~n::oftype(Integer) + ~~z) => trig_f(+(~~x..., ~~z...)) # reverse addition order
-       @rule trig_f(~~x + ~n::oftype(Integer) * ~y::multiple_of(2π) + ~~z) => trig_f(+(~~x..., ~~z...))]
-      for trig_f ∈ (sin, cos, tan)]...
-     
+     #  @rule trig_f(~~x + ~y::multiple_of(2π) * ~n::oftype(Integer) + ~~z) => trig_f(+(~~x..., ~~z...)) # reverse addition order
+     #  @rule trig_f(~~x + ~n::oftype(Integer) * ~y::multiple_of(2π) + ~~z) => trig_f(+(~~x..., ~~z...))]
+     # for trig_f ∈ (sin, cos, tan)]...
+
      @acrule(sin(~x)^2 + cos(~x)^2 => one(~x))
      @acrule(sin(~x)^2 + -1        => cos(~x)^2)
      @acrule(cos(~x)^2 + -1        => sin(~x)^2)
--- a/src/simplify.jl
+++ b/src/simplify.jl
@@ -1,7 +1,8 @@
 ##### Numeric simplification
 ### Predicates
 
-multiple_of(x, tol=1e-10) = y -> (y isa Number) && abs(y % x) < 1e-10
+# https://github.com/shashi/SymbolicUtils.jl/issues/23
+#multiple_of(x, tol=1e-10) = y -> (y isa Number) && abs(y % x) < 1e-10
 
 isnumber(x) = x isa Number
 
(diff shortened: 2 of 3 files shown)


MasonProtter:
I disagree with removing these identities. I'd say that it's the identities that are correct and the floating point math that fails to live up to the identities. If we can do better than floating point math, then I think we should if possible.

YingboMa:
Giving suboptimal results is far better than wrong results.

ChrisRackauckas:
Respecting math and respecting floating point math are two different things. IIRC the plan is to have different rule sets, and having math simplify vs floating point math simplify would be a good distinction, since different applications require different choices.

YingboMa:
We could define our own lazy `pi`, but that's pretty inconvenient to use.

YingboMa:
> I'd say that it's the identities that are correct and the floating point math that fails to live up to the identities.

I mean, mathematically
```julia
multiple_of(x, tol=1e-10) = y -> (y isa Number) && abs(y % x) < 1e-10
```
is wrong.

MasonProtter:
That said, your example in https://github.com/shashi/SymbolicUtils.jl/issues/23#issue-606820738 is fairly convincing that things can go wrong if the integer is sufficiently big. This is unfortunate because `eps(2pi) ~ 9e-16` whereas `1/typemax(Int) ~ 1e-19` so there's always 64 bit integers that are big enough to cause problems.

MasonProtter:
However, consider the fact that 
```julia
julia> eps(1e16)
2.0
```
so any result you get from `tan(x)` where `x ~ 1e16` is meaningless anyways:
```julia
julia> tan(1e16 + nextfloat(1e16))
4.524145195044381

julia> tan(1e16)
-1.2451734357184066
```

YingboMa:
I doubt that's any better
```julia
julia> tan(0)
0.0

julia> tan(0 + (2pi - 1e-16) * 1<<63)
0.2762097146903194
```

MasonProtter:
Yingbo, trigonometric functions applied to float64s that are that big are utterly meaningless. There is no "right answer" to `tan(0 + (2pi - 1e-16) * 1<<63)` other than promoting everything to BigFloat.

YingboMa:
I very much oppose defining non-equivalence relations on floating-point numbers in a symbolic library. If anything, we should define our own lazy transcendental numbers.

MasonProtter:
> If anything, we should define our own lazy transcendental numbers.

Sure, that'd be great. But it's also useful to be able to do simplifications on generic julia code that isn't written using arbitrary precision or transcendental numbers.

MasonProtter:
```julia
julia> (1e16 + 1) - 1e16
0.0
```
I don't think this is a convincing argument for removing our `+` simplification rules.

YingboMa:
There is a reason that
```julia
julia> tan(2pi)
-2.4492935982947064e-16
```
is not zero. Because `2pi` is a floating-point. We cannot just treat any floating point that's close enough to 2pi as 2pi.

YingboMa:
I think instead, we should add rules on `sinpi/cospi`, they are designed to do that
```julia
julia> sin(2pi)
-2.4492935982947064e-16

julia> sinpi(2)
0.0
```

YingboMa:
> I don't think this is a convincing argument for removing our + simplification rules.

I think they are quite fundamentally different. In the case of 2pi, we don't have the information to determine that

```julia
julia> using SymbolicUtils

julia> @vars x::Real
(x,)

julia> x - x
((-1 * x) + x)

julia> x + 2pi
(6.283185307179586 + x)
```

MasonProtter:
In what way are they different? 
```julia
julia> t = term(-, term(+, 1e16, 1.0), 1e16); showraw(t)
((1.0e16 + 1.0) - 1.0e16)

julia> @rule(~x + ~y - ~x => ~y)(t)
1.0

julia> 1e16 + 1 - 1e16
0.0
```
If a user puts something meaningless into a symbolic library and get's a different answer than when they put that meaningless input into julia, that's their fault, not ours.

YingboMa:
You know mathematically that
```julia
julia> using SymbolicUtils

julia> @vars x::Real
(x,)

julia> x - x
((-1 * x) + x)
```
is zero, but how do you know
```julia
julia> x + 2pi
(6.283185307179586 + x)
```
has two pi?

MasonProtter:
It's floating point arithmetic. You don't know anything. Everything has implicit errorbars. 

In particular you do **not** know that `x + y - x == y`.

YingboMa:
I know I can do `x + y - x == y` because I have the expression, but I don't know `6.283185307179586` is 2pi.

MasonProtter:
```julia
julia> let x = 1e16, y = 1
           @assert x == x
           @assert x === x
           x + y - x == y
       end
false
```

If a user is not comfortable with accumulating error, they should not use types which accumulate error.

YingboMa:
I am not talking about accumulating errors at all. It's quite obvious that
```julia
julia> 6 + sum(((i, x),)->x/10^i, enumerate(reverse(digits(283185307179586))))
6.283185307179586
```
is not 2pi.

shashi:
>  I'd say that it's the identities that are correct and the floating point math that fails to live up to the identities.

This is controversial enough that we shouldn't have it in this package. Also floating point errors are best left as they are rather than do this type of rounding which complicates analysis.

These rules would only make sense if we used a symbolic pi, I'm open to adding that, and other constants that people may want to simplify.

YingboMa:
I don't get what's the problem.

> I know I can do x + y - x == y because I have the expression, but I don't know 6.283185307179586 is 2pi.

I didn't say anything about the floating-point error. You can do that simplification because you know the expression.

MasonProtter:
Alright, let's remove this.

If we make an IRTools or Cassette pass, we can use it to replace instances of `pi` with a symbolic pi and then we can resolve these things exactly.

ChrisRackauckas:
I know @shashi hates it but this is why `Constant` exists: at some point I think you do need to take control of "symbolic constants" and force them to not do the Julia thing.

shashi:
> I know @shashi hates it but this is why Constant exists: at some point I think you do need to take control of "symbolic constants" and force them to not do the Julia thing.

Well the way to do this kind of constant is probably, `pi = Variable{Const{:pi}}(:pi)` and export that, and make it be a symbolic citizen. Using `Const` here to separate `@vars pi` in user code.

As far as I can tell the purpuse of MTK Constant is it lets you view julia numbers as subtype of Expression. Which is different.

MasonProtter:
Is there a reason that the lines here are commented out instead of deleted?

PR #27, opened 2020-04-25
interface tests

MasonProtter:
Basic tests to see that a different symbolic type can be made to play well with our simplification. I had to use a lot more than the basic infrastructure laid out in the README.

### DIFF
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -29,9 +29,8 @@ function Base.show(io::IO, r::Rule)
 end
 
 function (r::Rule)(term)
-    term = to_symbolic(term)
     m, rhs, dict = r.matcher, r.rhs, r._init_matches
-    (to_symbolic ∘ m)((term,), dict, (d, n) -> n == 1 ? (@timer "RHS" rhs(d)) : nothing)
+    m((term,), dict, (d, n) -> n == 1 ? (@timer "RHS" rhs(d)) : nothing)
 end
 
 """
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -29,8 +29,9 @@ function Base.show(io::IO, r::Rule)
 end
 
 function (r::Rule)(term)
+    term = to_symbolic(term)
     m, rhs, dict = r.matcher, r.rhs, r._init_matches
-    m((term,), dict, (d, n) -> n == 1 ? (@timer "RHS" rhs(d)) : nothing)
+    (to_symbolic ∘ m)((term,), dict, (d, n) -> n == 1 ? (@timer "RHS" rhs(d)) : nothing)
 end
 
 """
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -204,6 +204,11 @@ struct RuleSet <: AbstractRule
     rules::Vector{AbstractRule}
 end
 
+struct RuleRewriteError
+    rule
+    expr
+end
+
 function (r::RuleSet)(term; depth=typemax(Int))
     rules = r.rules
     # simplify the subexpressions
@@ -221,8 +226,7 @@ function (r::RuleSet)(term; depth=typemax(Int))
             expr′ = try
                 @timer(repr(rules[i]), rules[i](expr))
             catch err
-                show_rule_error(rules[i], expr)
-                rethrow()
+                throw(RuleRewriteError(rules[i], expr))
             end
             if expr′ === nothing
                 # this rule doesn't apply
@@ -237,10 +241,10 @@ function (r::RuleSet)(term; depth=typemax(Int))
     return expr # no rule applied
 end
 
-@noinline function show_rule_error(rule, expr)
-    msg = "\nFailed to apply rule $(rule) on expression "
-    msg *= sprint(io->showraw(io, expr))
-    Base.showerror(stderr, ErrorException(msg), backtrace()[1:min(100, end)])
+@noinline function Base.showerror(io::IO, err::RuleRewriteError)
+    msg = "Failed to apply rule $(err.rule) on expression "
+    msg *= sprint(io->showraw(io, err.expr))
+    print(io, msg)
 end
 
 function timerewrite(f)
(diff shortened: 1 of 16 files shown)


MasonProtter:
Should I add a section where I try to get it working using only the interface from the README and mark them with `@test_broken`?

shashi:
That is a cool test!

I think we should update the README with anything you found!!! I just wrote those based on what was needed for ModelingToolkit.

I don't get your question... What would be broken, for example?

test/interface.jl
```diff
@@ -0,0 +1,22 @@
...
+SymbolicUtils.arguments(ex::Expr) = ex.args[2:end]
+
+for f ∈ [:+, :-, :*, :/, :^]
+    @eval begin
+        Base.$f(x::Union{Expr, Symbol}, y::Number) = Expr(:call, $f, x, y) |> to_symbolic
+        Base.$f(x::Number, y::Union{Expr, Symbol}) = Expr(:call, $f, x, y) |> to_symbolic
+        Base.$f(x::Union{Expr, Symbol}, y::Union{Expr, Symbol}) = (Expr(:call, $f, x, y)) |> to_symbolic
+    end
+end
+SymbolicUtils.nameof(f::Symbol) = f
```
shashi, line 15: Why was this needed?

MasonProtter: You're right, it seems to not be needed.

test/interface.jl
```diff
@@ -0,0 +1,22 @@
...
+SymbolicUtils.symtype(::Union{Expr, Symbol}) = Number
+SymbolicUtils.istree(ex::Expr) = ex.head == :call
+SymbolicUtils.operation(ex::Expr) = ex.args[1]
+SymbolicUtils.arguments(ex::Expr) = ex.args[2:end]
+
+for f ∈ [:+, :-, :*, :/, :^]
+    @eval begin
+        Base.$f(x::Union{Expr, Symbol}, y::Number) = Expr(:call, $f, x, y) |> to_symbolic
+        Base.$f(x::Number, y::Union{Expr, Symbol}) = Expr(:call, $f, x, y) |> to_symbolic
+        Base.$f(x::Union{Expr, Symbol}, y::Union{Expr, Symbol}) = (Expr(:call, $f, x, y)) |> to_symbolic
```
MasonProtter, line 12: The main problem is that we need the `|>  to_symbolic` here.  If I get rid of it and try to do a simplification, this happens:
```julia
julia> ex = 1 + (:x - 2)
:((+)(1, (-)(x, 2)))

julia> SymbolicUtils.to_symbolic(ex)
((+)(x, -2) + 1)

julia> simplify(ex) |> dump
Term{Number}
  f: + (function of type typeof(+))
  arguments: Array{Any}((2,))
    1: Expr
      head: Symbol call
      args: Array{Any}((3,))
        1: + (function of type typeof(+))
        2: Symbol x
        3: Int64 -2
    2: Int64 1
```
Note that on the outside, this is a `Term`, but it's first argument is an `expr`
this is because the `@rule(-(~x, ~y) => ~x + -1(~y))` causes the operation `:x + -2` to happen which results in an `Expr`. I guess I can make it so that a `RuleSet` calls `to_symbolic` before every rule application instead of just once at the very start of `simplify`.

MasonProtter: Actually wait, we could just make the `matcher` system rely on `istree`, `operation` and `arguments` an then we might not even need to convert things to `Term`, right?

shashi: Hmm matcher can definitely work with those 3 functions! But the problem is you need to construct `tree`s when rewriting, so we do that with the `Term` constructor. I didn't want to complicate that a whole lot.

https://github.com/SciML/ModelingToolkit.jl/pull/326/files#diff-fd73d8caed68246d177d495cbfff183dR25

this was required. Maybe we should just document `to_symbolic` here. I think we should document `to_symbolic`.

So for conversion you either need to define 1) `to_symbolic` or 2) `istree`, `arg`, `op` combo. For expression trees 2) is preferred because it will automatically infer types of the terms.

MasonProtter: I might be misunderstanding you here, but defining `to_symbolic` is not enough, because when you apply rules you get objects which are no longer `Term`s.

shashi: Yeah `to_symbolic` is the function which converts any other IR into that of this package.
So if you see my link there is a `to_symbolic` method to convert to `SymbolicUtils.Variable`, you may need to define `to_symbolic(::Symbol)` here. `to_symbolic` should be the main part of the interface.

MasonProtter: Oh I see what you're saying now. I didn't see your comment until now, but I got everything working by just making `Rule`s apply `to_symbolic` before and after they're called which does work.

src/rule_dsl.jl
```diff
@@ -29,8 +29,9 @@ function Base.show(io::IO, r::Rule)
 end
 
 function (r::Rule)(term)
+    term = to_symbolic(term)
     m, rhs, dict = r.matcher, r.rhs, r._init_matches
-    m((term,), dict, (d, n) -> n == 1 ? (@timer "RHS" rhs(d)) : nothing)
+    (to_symbolic ∘ m)((term,), dict, (d, n) -> n == 1 ? (@timer "RHS" rhs(d)) : nothing)
```
shashi, line 34: why is this required? the default `to_symbolic` thing should recursively convert everything to this package's types.

src/types.jl
```diff
@@ -84,9 +84,7 @@ Base.one( s::Symbolic) = one( symtype(s))
 
 Base.zero(s::Symbolic) = zero(symtype(s))
 
-@noinline function promote_symtype(f, xs...)
-    error("promote_symtype($f, $(join(xs, ", "))) not defined")
-end
```
shashi, line 89: Are you sure about this?

MasonProtter: I think so. Otherwise, if your symtype is `Any`, you get things like `promote_symtype(+, Any, Int)` not defined which is really annoying. 

Simply promoting to `Any` seems like a sensible option

src/rule_dsl.jl
MasonProtter: reverted

src/types.jl
shashi: Okay makes sense!

PR #40, opened 2020-04-26
add shuffle rules

MasonProtter:
This adds a test-set where we apply this to a bunch of expressions:
```julia
R1 = RuleSet(SymbolicUtils.SIMPLIFY_RULES)
R2 = RuleSet(shuffle(R1.rules))
R1(ex) == R1(R2(ex))
```
the idea is that apply rules in a random order and then simplifying using the standard order should be the same as just simplifying with the standard order. 

This might be better suited to being done in the fuzzer, but I didn't really understand the fuzzer very well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,6 +18,7 @@ julia = "1"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [targets]
 test = ["Test", "Random"]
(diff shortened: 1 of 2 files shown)


shashi:
> the idea is that apply rules in a random order and then simplifying using the standard order should be the same as just simplifying with the standard order.

I'm not a 100% if this has to be true.

> This might be better suited to being done in the fuzzer, but I didn't really understand the fuzzer very well.

If you call `inputs = Set(); gen_rand_expr(inputs; min_depth=2)` you will get a random expression that you can test this way.

The fuzz.jl file also has some tests but we should refactor it.

PR #41, opened 2020-04-27
Add a fixpoint function, compute fixpoint in simplify

shashi:


MasonProtter:
Actually, I should say that personally, I'd prefer if the `fixpoint` behaviour was just a kwarg in `RuleSet`, e.g. `(R::RuleSet)(t; fixpoint=true)` but I don't care too much.

PR #42, opened 2020-04-28
Update tufte.css

tlienart:
Get the nice font


shashi:
You are awesome! thanks!

PR #44, opened 2020-04-29
add example to readme

MasonProtter:


shashi:
that last bit is deceptive, it's only showing 1 because it prints simplified :-p was that intended?

shashi:
I think it's ok to write

```
simplify(cos(y)^2 + sinh(im*y)^2, vcat(SymbolicUtils.SIMPLE_RULES, r))
```

shashi:
Oh actually I think it's ok since a previous example shows it's simplifying to print (or construct, in theory).

PR #47, opened 2020-04-30
Try to autodeploy website

shashi:
added `Pkg.instantiate()` to build


shashi:
@tlienart updates to a PR don't seem to trigger a rebuild, any ideas if that will be possible?

tlienart:
I think it’s possible, I’ll ask on discourse because I’m not very good with these GitHub actions 🙈

Issue #30, opened 2020-04-26
simplification of (a^m)^n


shashi:
The fuzzer found this interesting case:

```
Test failed for expression
    ((c ^ 44) ^ -66.20537843708486) = 0.0
Simplified to:
    (c ^ -2913.0366512317337) = Errored(DomainError(-2.517537638236398, "Exponentiation yielding a complex result requires a complex argument.\nReplace x^y with (x+0im)^y, Complex(x)^y, or similar."))
On inputs:
    Any[c] = [-2.517537638236398]
```

in general

```
(x^1/2)^2
```

is `+-x`

```
(x^2)^(1//2)
```
is `x`.

PR #51, opened 2020-05-01
Simplify rules refactor

MasonProtter:
Okay, this changes how we do simplification. Instead of one big, `RuleSet` which is applied, this uses nested rulesets to determine which rules actually get applied. In theory, this should be more efficient than our current system once we have more rules, but I find that currently it's a couple percent slower on this benchmark compared to master:
```julia
using SymbolicUtils, Random
Random.seed!(1)
let
    @syms a b c d
    expr1 = foldr((x,y)->rand([*, /, +])(x,y), rand([a,b,c,d,1,2], 20))
    SymbolicUtils.@timerewrite simplify(expr1)
end
```
(note that you have to run it like 3 times to get compilation out of the way).

I'll write up a more detailed report of what I changed and why in a bit.

### DIFF
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -257,8 +257,6 @@ end
 
 getdepth(::RuleSet) = typemax(Int)
 
-Base.vcat(Rs::RuleSet...) = RuleSet(vcat((R -> R.rules).(Rs)...))
-
 function fixpoint(f, x; kwargs...)
     x1 = f(x; kwargs...)
     while !isequal(x1, x)
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -214,8 +214,9 @@ a RuleSet until there are no changes.
 struct RuleSet <: AbstractRule
     rules::Vector{AbstractRule}
     applyall::Bool
+    recurse::Bool
 end
-RuleSet(rules; applyall=false) = RuleSet(rules, applyall)
+RuleSet(rules; applyall=false, recurse=true) = RuleSet(rules, applyall, recurse)
 
 
 struct RuleRewriteError
@@ -231,7 +232,7 @@ function (r::RuleSet)(@nospecialize(term); depth=typemax(Int))
         return term
     end
     if term isa Symbolic
-        if term isa Term
+        if term isa Term && r.recurse
             expr = Term{symtype(term)}(operation(term),
                                        map(t -> r(t, depth=depth-1), arguments(term)))
         else
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -213,10 +213,7 @@ a RuleSet until there are no changes.
 """
 struct RuleSet <: AbstractRule
     rules::Vector{AbstractRule}
-    applyall::Bool
-    recurse::Bool
 end
-RuleSet(rules; applyall=false, recurse=true) = RuleSet(rules, applyall, recurse)
 
 
 struct RuleRewriteError
@@ -224,7 +221,7 @@ struct RuleRewriteError
     expr
 end
 
-function (r::RuleSet)(@nospecialize(term); depth=typemax(Int))
+function (r::RuleSet)(@nospecialize(term); depth=typemax(Int), applyall=false, recurse=true)
     rules = r.rules
     term = to_symbolic(term)
     # simplify the subexpressions
@@ -232,7 +229,7 @@ function (r::RuleSet)(@nospecialize(term); depth=typemax(Int))
         return term
     end
     if term isa Symbolic
-        if term isa Term && r.recurse
+        if term isa Term && recurse
             expr = Term{symtype(term)}(operation(term),
                                        map(t -> r(t, depth=depth-1), arguments(term)))
         else
@@ -249,7 +246,7 @@ function (r::RuleSet)(@nospecialize(term); depth=typemax(Int))
                 continue
             else
                 expr = r(expr′, depth=getdepth(rules[i]))# levels touched
-                r.applyall || return expr
+                applyall || return expr
             end
         end
     else
@@ -262,16 +259,16 @@ getdepth(::RuleSet) = typemax(Int)
 
 Base.vcat(Rs::RuleSet...) = RuleSet(vcat((R -> R.rules).(Rs)...))
 
-function fixpoint(f, x)
-    x1 = f(x)
+function fixpoint(f, x; kwargs...)
+    x1 = f(x; kwargs...)
     while !isequal(x1, x)
         x = x1
-        x1 = f(x)
+        x1 = f(x; kwargs...)
     end
     return x1
 end
 
-fixpoint(f) = x->fixpoint(f, x)
+fixpoint(f; kwargs...) = x -> fixpoint(f, x; kwargs...)
 
 @noinline function Base.showerror(io::IO, err::RuleRewriteError)
     msg = "Failed to apply rule $(err.rule) on expression "
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -214,8 +214,9 @@ a RuleSet until there are no changes.
 struct RuleSet <: AbstractRule
     rules::Vector{AbstractRule}
     applyall::Bool
+    recurse::Bool
 end
-RuleSet(rules; applyall=false) = RuleSet(rules, applyall)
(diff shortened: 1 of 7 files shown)


src/rulesets.jl
```diff
@@ -1,62 +1,86 @@
...
+    @acrule((~x)^(~n) * (~x)^(~m) => (~x)^(~n + ~m))
+
+    @acrule((~z::_isone  * ~x) => ~x)
+    @acrule((~z::_iszero *  ~x) => ~z)
+    @rule(*(~x) => ~x)
+], recurse=false)
+
+const POW_RULES = RuleSet([
+    @rule(^(*(~~x), ~y) => *(map(a->pow(a, ~y), ~~x)...))
+    @rule((((~x)^(~p))^(~q)) => (~x)^((~p)*(~q)))
```
shashi, line 59: I think we should remove this, or restrict it to integers. But you don't have to do it in this PR.

see #30

src/rulesets.jl
```diff
@@ -1,62 +1,86 @@
...
+#     @acrule(tan(~x)^2 + -1*sec(~x)^2 => one(~x))
+#     @acrule(tan(~x)^2 +  1 => sec(~x)^2)
+#     @acrule(sec(~x)^2 + -1 => tan(~x)^2)
+
+#     @acrule(cot(~x)^2 + -1*csc(~x)^2 => one(~x))
+#     @acrule(cot(~x)^2 +  1 => csc(~x)^2)
+#     @acrule(csc(~x)^2 + -1 => cot(~x)^2)
+# ])
 
-SIMPLIFY_RULES = vcat(BASIC_NUMBER_RULES, TRIG_RULES)
```
shashi, line 59: We mention this a couple of times in the docs and manual. Should update.

src/simplify.jl
```diff
@@ -8,19 +8,29 @@ these rules are `SymbolicUtils.SIMPLIFY_RULES`. If `fixpoint=true`
...
 Applies them once if `fixpoint=false`.
 """
-function simplify(x, rules=SIMPLIFY_RULES; fixpoint=true)
+function simplify(x, rules::Vector; fixpoint=true)
     if fixpoint
         SymbolicUtils.fixpoint(RuleSet(rules))(x)
     else
         RuleSet(rules)(x)
     end
 end
```
shashi, line 17: did you leave this for backwards compatibility? Should probably call the below method directly.

test/interface.jl
```diff
@@ -17,14 +17,15 @@ end
 ex = 1 + (:x - 2)
 
 @test to_symbolic(ex) == Term{Any}(+, [1, Term{Any}(-, [Sym{Any}(:x), 2])])
+@test simplify(ex) == to_symbolic(ex) # Not simplified because symtype Any
```
shashi, line 20: nice!

src/rule_dsl.jl
```diff
@@ -212,22 +213,26 @@ a RuleSet until there are no changes.
...
+function (r::RuleSet)(@nospecialize(term); depth=typemax(Int))
     rules = r.rules
     term = to_symbolic(term)
     # simplify the subexpressions
     if depth == 0
         return term
     end
     if term isa Symbolic
-        if term isa Term
+        if term isa Term && r.recurse
```
shashi, line 235: why would you not want to recurse?

shashi: Oh I see what's up. So you want to recurse only from the topmost RuleSet

src/rulesets.jl
```diff
@@ -1,62 +1,86 @@
...
+const SIMPLIFY_RULES = RuleSet([
+    @rule ~t::sym_isa(Number) => NUMBER_RULES(~t)
+], applyall=true)
+
+const NUMBER_RULES = RuleSet([
+    @rule ~t => ASSORTED_RULES(~t)
+    @rule ~t::is_operation(+) => PLUS_RULES(~t)
+    @rule ~t::is_operation(*) => TIMES_RULES(~t)
+    @rule ~t::is_operation(^) => POW_RULES(~t)
+], applyall=true)
```
shashi, line 11: Why not set `recurse=false` here?

src/rule_dsl.jl
shashi: But it does not seem like it should be a flag on the RuleSet type but should be an internal argument to this function. It should start off with true and become false if going into a nested RuleSet.

shashi: That would leave the `RuleSet` objects free to be composed however without worrying which one should go on top. Tell me if I'm talking nonsense!

src/rule_dsl.jl
```diff
@@ -243,7 +248,8 @@ function (r::RuleSet)(term; depth=typemax(Int))
                 # this rule doesn't apply
                 continue
             else
-                return r(expr′, depth=getdepth(rules[i])) # levels touched
+                expr = r(expr′, depth=getdepth(rules[i]))# levels touched
+                r.applyall || return expr
```
shashi, line 252: So does this say if there's one change go to the parent `RuleSet`? Seems like the same amount of info as recurse :) and sure enough in your code you either set recurse or this and not both. Why not use `!r.recurse` ? Better yet, make it an optional state of this function!

src/rulesets.jl
```diff
@@ -1,62 +1,86 @@
...
+#     @acrule(cos(~x)^2 + -1        => sin(~x)^2)
+
+#     @acrule(tan(~x)^2 + -1*sec(~x)^2 => one(~x))
+#     @acrule(tan(~x)^2 +  1 => sec(~x)^2)
+#     @acrule(sec(~x)^2 + -1 => tan(~x)^2)
+
+#     @acrule(cot(~x)^2 + -1*csc(~x)^2 => one(~x))
+#     @acrule(cot(~x)^2 +  1 => csc(~x)^2)
+#     @acrule(csc(~x)^2 + -1 => cot(~x)^2)
+# ])
```
shashi, line 82: Did you mean to put these back in later?

src/simplify.jl
MasonProtter: yeah, makes sense.

src/rule_dsl.jl
MasonProtter: Yeah, you're tight, I'll change this.

src/rule_dsl.jl
MasonProtter: Yeah, it should be an option for the function instead of the ruleset. It is different than `recurse`. If you have `RuleSet([r1, r2])` and `r1` always matches, you'll never use `r2` if you don't turn on `applyall`.

src/rulesets.jl
MasonProtter: Unfortunately, the way things work out, you need both `SIMPLIFY_RULES` and `NUMERIC_RULES` to recurse.

MasonProtter:
Okay, this is now a bit over 12% faster than master on my benchmark above (80ms -> 70ms) on my machine after running a few times.

MasonProtter:
I'll add some tests to improve coverage.

shashi:
That :x: mark for coverage is annoying indeed

src/rule_dsl.jl
shashi: no matter what `applyall` is, this line just leads to `return expr` here or below. Am I reading it wrong?

src/rulesets.jl
```diff
@@ -1,14 +1,15 @@
 
 const SIMPLIFY_RULES = RuleSet([
-    @rule ~t::sym_isa(Number) => NUMBER_RULES(~t)
-], applyall=true)
+    @rule ~t::sym_isa(Number) => NUMBER_RULES(~t, applyall=true, recurse=true)
```
shashi, line 3: that's nice!

shashi:
> I'll write up a more detailed report of what I changed and why in a bit.

It would be nice to document `applyall` and `recurse` keywords more. Maybe in the docstring itself and give a couple of examples of the behavior. It took me a while to get what they do.

shashi:
I see why it makes sense for the RuleSets to have some flags, sorry for pushing back on that :sweat_smile: I do think `applyall` belongs there. Maybe we should call it `flatten`.

`recurse` seems like an internal optimization kinda thing so we don't recurse the recursion while it recurses the recursion. Am I right?

Just a note: it seems like we are manually creating a suffix tree of predicates. Maybe this can be auto-generated in the future :thinking:

PR #55, opened 2020-05-01
powers of sums

JeffreySarnoff:
```
=> (a + b)^(2:5)
=> (a - b)^(2:5)                 # ?
=> (a + b + c)^(2:3)
=> (a + b - c)^(2:3)             # ?
=> (a - b - c)^(2:3)             # ?
```
are the rules marked `# ?` needed or do other rules handle (-(x))?

### DIFF
--- a/src/rulesets.jl
+++ b/src/rulesets.jl
@@ -4,11 +4,12 @@ const SIMPLIFY_RULES = RuleSet([
 ])
 
 const NUMBER_RULES = RuleSet([
-    @rule ~t               => ASSORTED_RULES(~t, recurse=false)
-    @rule ~t::is_operation(+) =>  PLUS_RULES(~t, recurse=false)
-    @rule ~t::is_operation(*) => TIMES_RULES(~t, recurse=false)
-    @rule ~t::is_operation(^) =>   POW_RULES(~t, recurse=false)
-    @rule ~t                  =>  TRIG_RULES(~t, recurse=false)
+    @rule ~t                => ASSORTED_RULES(~t, recurse=false)
+    @rule ~t::is_operation(+) =>   PLUS_RULES(~t, recurse=false)
+    @rule ~t::is_operation(*) =>  TIMES_RULES(~t, recurse=false)
+    @rule ~t::is_operation(^) =>    POW_RULES(~t, recurse=false)
+    @rule ~t::is_operation(^) => POWSUM_RULES(~t, recurse=false)
+    @rule ~t                  =>   TRIG_RULES(~t, recurse=false)
 ])
 
 const PLUS_RULES = RuleSet([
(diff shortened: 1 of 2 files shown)


src/rulesets.jl
```diff
@@ -4,11 +4,12 @@ const SIMPLIFY_RULES = RuleSet([
...
-    @rule ~t               => ASSORTED_RULES(~t, recurse=false)
-    @rule ~t::is_operation(+) =>  PLUS_RULES(~t, recurse=false)
-    @rule ~t::is_operation(*) => TIMES_RULES(~t, recurse=false)
-    @rule ~t::is_operation(^) =>   POW_RULES(~t, recurse=false)
-    @rule ~t                  =>  TRIG_RULES(~t, recurse=false)
+    @rule ~t                => ASSORTED_RULES(~t, recurse=false)
+    @rule ~t::is_operation(+) =>   PLUS_RULES(~t, recurse=false)
+    @rule ~t::is_operation(*) =>  TIMES_RULES(~t, recurse=false)
+    @rule ~t::is_operation(^) =>    POW_RULES(~t, recurse=false)
+    @rule ~t::is_operation(^) => POWSUM_RULES(~t, recurse=false)
```
MasonProtter, line 11: I think this should be
```suggestion
    @rule ~t::is_operation(+) => POWSUM_RULES(~t, recurse=false)
```
The `is_operation(f)` function tests if a term's outermost `operation` is equal to `f`. For your rules, this will be `+` instead of `^`.

src/rulesets.jl
```diff
@@ -50,6 +51,28 @@ const POW_RULES = RuleSet([
...
+    @acrule( (~x)^5 - 5*(~x)^4*(~y) + 10*(~x)^3*(~y)^2 - 10*(~x)^2*(~y)^3 + 5*(~x)*(~y)^4 - (~y)^5 => ((~x) - (~y))^5 )
+
+    @acrule( (~x)^2 + 2*(~x)*(~y) + 2*(~x)*(~z) + (~y)^2 + 2*(~y)*(~z) + (~z)^2 => ((~x) + (~y) + (~z))^2 )
+    @acrule( (~x)^2 + 2*(~x)*(~y) - 2*(~x)*(~z) + (~y)^2 - 2*(~y)*(~z) + (~z)^2 => ((~x) + (~y) - (~z))^2 )
+    @acrule( (~x)^2 - 2*(~x)*(~y) - 2*(~x)*(~z) + (~y)^2 + 2*(~y)*(~z) + (~z)^2 => ((~x) - (~y) - (~z))^2 )
+
+    @acrule( (~x)^3 + 3*(~x)^2*(~y) + 3*(~x)^2*(~z) + 3*(~x)*(~y)^2 + 6*(~x)*(~y)*(~z) + 3*(~x)*(~z)^2 + (~y)^3 + 3*(~y)^2*(~z) + 3*(~y)*(~z)^2 + (~z)^3 => ((~x) + (~y) + (~z))^3 )
+    @acrule( (~x)^3 + 3*(~x)^2*(~y) - 3*(~x)^2*(~z) + 3*(~x)*(~y)^2 - 6*(~x)*(~y)*(~z) + 3*(~x)*(~z)^2 + (~y)^3 - 3*(~y)^2*(~z) + 3*(~y)*(~z)^2 - (~z)^3 => ((~x) + (~y) - (~z))^3 )
+    @acrule( (~x)^3 - 3*(~x)^2*(~y) - 3*(~x)^2*(~z) + 3*(~x)*(~y)^2 + 6*(~x)*(~y)*(~z) + 3*(~x)*(~z)^2 - (~y)^3 - 3*(~y)^2*(~z) - 3*(~y)*(~z)^2 - (~z)^3 => ((~x) - (~y) - (~z))^3 )
+])
```
MasonProtter, line 74: In the `ASSORTED_RULES`, we transform things like `x - y` into `x + -1 * y`, so I don't think it helps to have the `-` signs in here. Instead, I think 
```julia
@acrule( (~x)^2 - 2*(~x)*(~y) + (~y)^2 => ((~x) - (~y))^2 )
```
should read
```julia
@acrule( (~x)^2 + (-2)*(~x)*(~y) + (~y)^2 => ((~x) + -(~y))^2 )
```

MasonProtter:
Thanks for writing these out! I think with some tweaks these could be useful rules.

JeffreySarnoff:
happy to oblige

MasonProtter:
I'm a little worried about the performance impact, because if `simplify` goes into this, there's a *lot* of match tests to do be done. I would feel better if we could come up with a good way to test at the top level if these rules should be applied. 

Maybe we could do something like this:
```julia
isa_pow_term(@nospecialize(t)) = t isa Term && operation(t) == (^) 
function should_apply_powsum_rules(@nospecialize(t))
    (t isa Term) && (operation(t) == (+)) && count(isa_pow_term, arguments(t)) > 2
end
```
that way we only go into this ruleset if the `operation` is `+` and if there are at least two subterms whose operation is `^`?

shashi:
>  performance impact

we can only call these rules if number of args to + have length >=3, for example. Or we can maybe build in that optimization into the general matcher -- we know that for these AC rules, there need to be at least N number of arguments to +.

shashi:
I'm trying this out locally, doesn't seem to work.

I'm concerned about this case:

`a^2 + b^2 + 2a*b - 2a*b - a^2`

These rules might make that expression really complicated and leave them that way.

shashi:
scmutils does the opposite,

```julia

2 error> (* (+ 'a 'b) (- 'a 'b))
#|
(+ (expt a 2) (* -1 (expt b 2)))
|#

2 error> (* (+ 'a 'b) (- 'a 'c))
#|
(+ (expt a 2) (* a b) (* -1 a c) (* -1 b c))
|#

2 error> (* (+ 'a 'b) (+ 'a 'b))
#|
(+ (expt a 2) (* 2 a b) (expt b 2))
|#
```

I'm not looking forward to adding these rules to the default `simplify` because it is not clear if this plays well with the other rules, and it adds overhead to the base simplify.

That said I think these should be an optional RuleSet! We should start a "Standard library" of rules. (One file is too less.) And should test and document the variety of available rules.

JeffreySarnoff:
I had hoped the plumbing let the shorter of these rules be applied without any possibility of running in circles or getting stuck in corners, and would get through them quickly.  Clearly, this is not the case at present.  Rather than consider this PR for immediate consideration, I would be happy to use this thread to clarify what sort of rules are most appropriate to be considered for inclusion with the other basic simplifying rules and how to identify other cases -- and how best to utilize them where appropriate without foot-shooting performance or outcomes.

shashi:
> I had hoped the plumbing let the shorter of these rules be applied without any possibility of running in circles or getting stuck in corners

Yes currently we don't have a metric for what "simpler" means. If we did that it would be possible to assign a score to all branches of simplification possible and take the best one. But "simpler" depends on application (e.g. if we want to generate Julia code from the expression we want to minimize number of operations, but that may not be simpler for humans). We need to think about it more.

JeffreySarnoff:
[Ten commandments for good default expression simplification by David R. Stoutemyer](https://www.sciencedirect.com/science/article/pii/S0747717110001471)

MasonProtter:
I'm a little surprised by this behaviour in scmutils. I'd argue that it makes a lot of sense to turn `x^2 + 2x*y + y^2` into `(x + y)^2`. That's a much more compact expression visually as well as requiring less operations for the computer to execute if `x` and `y` were substituted with numbers. 

For what it's worth, Mathematica simplifies along these lines.

JeffreySarnoff:
This work and the works of others who will utilize this as a foundational constituent should rely on some intuitively coherent approach to simplification and the partial ordering of expressions of greater/lesser/similar simplicity.
```
(a + b)^2          # one constant utilized, one occurence of utilized constants
                   # two symbols utilized, two occurrences of utilized symbols
                   # one use of '±', one use of '^'

a^2 + b^2 + 2*a*b  # one constant utilized, three occurrences of utilized constants
                   # two symbols utilized, four occurrences of utilized symbols
                   # two uses of '±', two uses of '*', two uses of '^'
```

shashi:
> we can only call these rules if number of args to + have length >=3, for example. Or we can maybe build in that optimization into the general matcher -- we know that for these AC rules, there need to be at least N number of arguments to +.

Cool, I was trying to add this and found out that whole codepath is not taken because `permutations([1,2], 3)` is an empty iterator. So ACRule already has these smarts!

shashi:
We just need to reduce the overhead of adding new rules. Now that we have a hierarchy of rules, we could make some sub rulesets (at the leaf level) be `RuleSet{T}(T::Tuple)` and use recursion or generated function to facilitate really tight code.

But I tried this over the weekend (on `ASSORTED_RULES`) and tuple-recursion, it kills performance though (half a second to TEN seconds for the 1000 term foldr simplify).

shashi:
I now think the right way to implement this is to have a polynomial factorization step. I've added support for using AbstractAlgebra.jl's multivariate polynomial library to "normalize" or expand any expression. So if anyone implements a factorization routine for those, we can immediately start using that. Should it be default behavior? I don't know but the important thing is it should become possible in due time.

linwaytin:
I think maybe we should consider `simplify` and `factor` as two separate functions.
In `Sympy` and `Maxima`, they have `simplify` to do some basic simplifications, and `factor` is focused on factorize an expression into a product of irreducible expressions.
In my opinion, factorization in this sense is a more well-defined problem.
On the other hand, simplification is sort of vague. As mentioned above, what is the metric for simplification?
Simplification is also more than just factorization. 
For example, I would say `x^2 + 2xy + y^2 + 1` can be simplified to `(x+y)^2 + 1`, although the this expression is not factorizable.

I would suggest that we consider factorization as another issue, and focus on simple powers of sums, which should be useful by themselves.
However, if the rule set becomes bigger and bigger,   as mentioned above the performance will be a problem.
Also, in the code the rules are listed to n=5 or so.
How about higher order rules are needed?
As mentioned by @MasonProtter , there should be some check before applying the rules.
I also think this should be the general scheme for the function `simplify`.
That should mitigate the performance problem when the rule set is huge.

shashi:
Was just talking to @sumiya11 who is the author of Groebner.jl -- he is told that a good factorization takes a couple of years to implement haha. I guess we could create a package which calls Flint for now!?

I agree that `factorize` should be separate from `simplify` but `simplify` should call `factorize` if it turns out to be beneficial, or have an option to do so.

PR #75, opened 2020-05-07
boolean methods and rules

### LINKED ISSUE #67: Symbolic{Bool} and simplification
Allow symbolic conditions in rules:

```julia
@rule abs(~x) => ~x >= 0 ? ~x : -~x
# In a Bool ruleset:
@rule ~x >= 0 => isnonnegative(~x, @CTX) # isnonnegative returns Union{Bool, Nothing}
```

This should be lowered such that if the truth value of `~x >= 0` cannot be ascertained, then it would leave the input expression unchanged.

We can potentially allow expressions with if conditions. Maybe just by adding an `ifelse` method.

shashi:
Added some boolean algebra rules.

`cond` is like `ifelse`. Unfortunately `ifelse` itself is not a generic function.

```julia
julia>@syms a::Real
julia> simplify(substitute(cond(a < 0, -a,a), Dict(a=>-1)))
-1
```

closes #67

### DIFF
--- a/src/SymbolicUtils.jl
+++ b/src/SymbolicUtils.jl
@@ -33,6 +33,7 @@ include("types.jl")
 
 
 using SpecialFunctions, NaNMath
+export cond
 include("methods.jl")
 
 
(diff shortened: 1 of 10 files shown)


shashi:
I added a couple of comments about what they do.

PR #84, opened 2020-05-09
Descendants

MasonProtter:
This makes `Term`s carry around their number of descendants. It seems to offer no real performance improvement from what I see locally though. Maybe I did something wrong?

### DIFF
--- a/src/rewriters.jl
+++ b/src/rewriters.jl
@@ -167,7 +167,7 @@ function (p::Walk{ord, C, true})(x) where {ord, C}
         end
         if istree(x)
             _args = map(arguments(x)) do arg
-                if node_count(arg) > p.thread_cutoff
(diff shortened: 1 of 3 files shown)


shashi:
looks like we need to handle the empty case either way.

YingboMa:
Rebase?

shashi:
There are regressions, I guess we wastefully compute this too many times and throw it away... It might be useful to do some type of occursin_info kind of thing when necessary here (at the top level when threading, mayybe).

YingboMa:
yeah, that’s a good idea.

On July 14, 2020 at 6:34 GMT, Shashi Gowda <notifications@github.com> wrote:

There are regressions, I guess we wastefully compute this too many times and throw it away... It might be useful to do some type of occursin_info kind of thing when necessary here (at the top level when threading, mayybe).

Issue #94, opened 2020-05-19
Simplification seems to stop?


ChrisRackauckas:
```julia
using ModelingToolkit, StaticArrays, LinearAlgebra
using DiffEqBase

# Define some variables
@parameters t σ ρ β
@variables x(t) y(t) z(t)
@derivatives D'~t

# Define a differential equation
eqs = [D(x) ~ σ*(y-x),
       D(y) ~ -z-y,
       D(z) ~ y - β*z]

sys = ODESystem(eqs)
isequal(getfield.(sys.eqs,:rhs),calculate_jacobian(sys)*[x,y,z])
@show simplify.(getfield.(sys.eqs,:rhs) - calculate_jacobian(sys)*[x,y,z])

3-element Array{Expression,1}:
 σ * (y(t) + -1 * x(t)) + -1 * (y(t) * σ + -1 * x(t) * σ)
                  -1 * (y(t) + z(t) + -1 * (y(t) + z(t)))
                              ModelingToolkit.Constant(0)
```

That shouldn't be the most difficult case, so I wonder if the stopping criterion is off?

shashi:
We don't have rules to distribute scalar multiplication, we attempt a bit to do the reverse, but clearly not effective enough.

So I'll fix it by adding a rule to distribute.

shashi:
Keeping this open till release.

This normalized simplify most likely should be another function.

shashi:
These should simplify on calling `to_mtk(polynormalize(to_symbolic(x)))`.

PR #95, opened 2020-05-19
Number rules fixes

shashi:
Solves #94 by distributing * over +.

### DIFF
--- a/src/rule_dsl.jl
+++ b/src/rule_dsl.jl
@@ -185,8 +185,8 @@ getdepth(r::ACRule) = getdepth(r.rule)
 macro acrule(expr)
     arity = length(expr.args[2].args[2:end])
     quote
-        ACRule(@rule($expr), $arity)
-    end
+        SymbolicUtils.ACRule(SymbolicUtils.@rule($(expr)), $arity)
+    end |> esc
 end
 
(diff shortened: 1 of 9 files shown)


src/rule_dsl.jl
```diff
@@ -185,8 +185,8 @@ getdepth(r::ACRule) = getdepth(r.rule)
 macro acrule(expr)
     arity = length(expr.args[2].args[2:end])
     quote
-        ACRule(@rule($expr), $arity)
-    end
+        SymbolicUtils.ACRule(SymbolicUtils.@rule($(expr)), $arity)
```
asinghvi17, line 188:
```suggestion
        $(@__MODULE__).ACRule($(@__MODULE__).@rule($(expr)), $arity)
```
If I call this from Main, and SymbolicUtils is not loaded (e.g. when using ModelingToolkit) then explicitly referencing `SymbolicUtils` might fail.

shashi: thanks!

shashi:
I think we can fix this without using distribute.

```julia
julia> a * (b + -1*c) + -1*(b*a + -1 * c * a)
0
```
this already works, this is not the right fix.

Will close this and cherry pick into a new PR.

PR #105, opened 2020-06-12
in RuleSet loop handle cases where rules are no-op

shashi:
This gives a 2x speedup on

```
x2=foldl((x,y)->rand([*,/])(x,y),rand([a,b,c,d,1//2,2], 1000));
```

cc @MasonProtter


YingboMa:
It seems to introduce some overhead in https://travis-ci.com/github/JuliaSymbolics/SymbolicUtils.jl/jobs/348079067#L686

It can probably be solved by making a single rule application return `has_applied::Bool`.

shashi:
weird.....

PR #106, opened 2020-06-15
Rewriters

### LINKED ISSUE #101: Get Revise to work with rulesets.jl
Changing the rules does not revise, and prints a lot of nonsense errors.

shashi:
A rewriter is any function which takes an expression and returns an expression
or `nothing`. If `nothing` is returned that means there was no changes applicable
to the input expression.

The `SymbolicUtils.Rewriters` module contains some types which create and transform
rewriters.

- `Empty()` is a rewriter which always returns `nothing`
- `Chain(itr)` takes an iterator of rewriters returns a rewriter which applies each
   rewriter in `itr`. `nothing` is treated as no change.
- `RestartedChain(itr)` like `Chain` but restarts on the first successful application of
   one of the chained rewriters.
- `Prewalk(rw; threaded=false, thread_cutoff=100)` returns a rewriter which does a pre-order
   traversal of a given expression and applies the rewriter `rw`. `threaded=true` will
   use multi threading for traversal. `thread_cutoff` is the minimum number of nodes
   in a subtree which should be walked in a threaded spawn.
- `Postwalk(rw; threaded=false, thread_cutoff=100)` similarly does post-order traversal.
- `IfElse(cond, rw1, rw2)` runs the `cond` function on the input, applies `rw1` if cond
   returns true, `rw2` if it retuns false
- `If(cond, rw)` is the same as `IfElse(cond, rw, Empty())`
- `PassThrough(rw)` returns a rewriter which if `rw(x)` returns `nothing` will instead
   return `x` otherwise will return `rw(x)`.

Using this in place of RuleSet gives way more fine-grained control over sequence of rule applications, ability to Fixpoint or thread any sub-part of the rule application process.

Removes the context system added by #68 , it seems the same purpose can well be served by using rules that close over the contexts.

Also fixes #101 

Is faster than the current simplify (especially if random_term is benchmarked with `[*, /]` it's about 4x faster. for `[+, *]` it's like 5% faster.)


shashi:
For some reason this is even better on Julia nightly...

```julia
	  "simplify" => 7-element BenchmarkTools.BenchmarkGroup:
		  tags: []
		  "randterm (/, *):serial" => TrialJudgement(-87.87% => improvement)
		  "noop:Int" => TrialJudgement(-59.94% => improvement)
		  "noop:Sym" => TrialJudgement(-99.21% => improvement)
		  "randterm (/, *):thread" => TrialJudgement(-86.40% => improvement)
		  "noop:Term" => TrialJudgement(+2.66% => invariant)
		  "randterm (+, *):serial" => TrialJudgement(-12.94% => improvement)
		  "randterm (+, *):thread" => TrialJudgement(-5.16% => improvement)
```

PR #112, opened 2020-07-07
apply ASSORTED_RULES in a pre-order walk

shashi:
Helps greatly when simplifying `expand_derivatives` from ModelingToolkit. (23 sec vs, 8 sec)


shashi:
not worth the regression

PR #115, opened 2020-07-13
Remove unbound parameter

### LINKED ISSUE #114: Unbound args
```
pkg> add SymbolicUtils#master

julia> using SymbolicUtils
[ Info: Precompiling SymbolicUtils [d1185830-fcd6-423d-90d6-eec64667417b]

julia> using Test; detect_unbound_args(SymbolicUtils)
Skipping SymbolicUtils.@arule
Skipping SymbolicUtils.@fun
[1] (a::AbstractAlgebra.Generic.MPoly{T})(vals::U...) where {T<:Union{AbstractAlgebra.RingElem, AbstractFloat, Integer, Rational}, U<:Union{AbstractFloat, Integer, Rational}} in AbstractAlgebra.Generic at /home/graham/.julia/packages/AbstractAlgebra/NyIUy/src/generic/MPoly.jl:3758
[2] (acr::SymbolicUtils.ACRule)(term) where comm in SymbolicUtils at /home/graham/.julia/packages/SymbolicUtils/Qo6yr/src/rule.jl:308
```

Apparently this is can lead to performance hits (https://discourse.julialang.org/t/unused-where-t-causes-a-function-to-become-very-slow/39727) and it leads to failed tests for downstream packages that test `isempty(detect_unbound_args(MyPkgImportingSymbolicUtils))`

grahamas:
Fixes #114 

Tiny change to fix unbound parameter.

Still fails `isempty(detect_unbound_args(SymbolicUtils))` due to weird unbound parameter in AbstractAlgebra.

Might want to include explict test.  I've included a commit with that test (`@test_broken`, for now, and probably for the foreseeable future, due to the dependence on upstream compliance), but you can easily remove that commit if desired.

### DIFF
--- a/src/rule.jl
+++ b/src/rule.jl
@@ -304,7 +304,7 @@ end
 
 Base.show(io::IO, acr::ACRule) = print(io, "ACRule(", acr.rule, ")")
 
-function (acr::ACRule)(term) where {comm}
+function (acr::ACRule)(term)
     r = Rule(acr)
     if !(term isa Term)
         r(term)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -13,6 +13,7 @@ end
 SymbolicUtils.show_simplified[] = false
 
 #using SymbolicUtils: Rule
+@test_broken isempty(detect_unbound_args(SymbolicUtils))
 
 include("basics.jl")
 include("order.jl")


YingboMa:
Thanks!

PR #122, opened 2020-08-16
WIP: metadata system

shashi:
`Term` and `Sym` should maintain a dictionary which given a type will give you a single value which is the metadata associated with the object for that "context" type.

I will try to optimize this with `ImmutableDict`, but this is the general idea I'm going for.

Any comments @MasonProtter @YingboMa @ChrisRackauckas ?

Something like this is required to store array shapes to implement symbolic arrays.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -125,9 +125,11 @@ means the variable is a function with the type signature X -> Y where
 """
 struct Sym{T} <: Symbolic{T}
     name::Symbol
+    metadata::Dict{Type, Any}
 end
 
 const Variable = Sym # old name
+Sym{T}(x) where {T} = Sym{T}(x, Dict{Type,Any}())
 Sym(x) = Sym{symtype(x)}(x)
 
 Base.nameof(v::Sym) = v.name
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Making variables suddenly not bitstypes could have a large effect on performance. I am not sure about this one.

shashi:
It can be optimized. We can keep it bits in theory by using `ImmutableDict` and parameterizing by metadata type. We should wait to see some benchmarks.

shashi:
Benchmarks for this gets like 8-9% slower. without any optimizations whatsoever.

```
                  "randterm (/, *):thread" => TrialJudgement(+8.66% => regression)
                  "randterm (+, *):serial" => TrialJudgement(+8.68% => regression)
```

ChrisRackauckas:
What about something like in the context of the ReactionSystem -> ODESystem construction?

shashi:
> What about something like in the context of the ReactionSystem -> ODESystem construction?

Not sure what you're asking exactly, but the point of this dict is that many packages can add different metadata to symbols and use them in rules without conflicting with each others' metadata. So yeah if ReactionSystem -> ODESystem involves propagating metadata, you should be able to use it.

shashi:
Closing in favor of #215

Issue #134, opened 2020-10-29
Non-number symtypes on v0.6


david-pl:
When using symbolic types that are not `<:Number` on `v0.6.0`, the new `assert_number` check keeps you from easily combining such a type with numbers, e.g. the following will error:
```julia
using SymbolicUtils
struct QNumber end
@syms x::QNumber
2*x
```
What is the recommended way of dealing with this? Should I add a method to `assert_number`, e.g. `assert_number(::Symbolic{<:QNumber})=true` ?

shashi:
Would it make sense in your case to have `struct QNumber <: Number end`? I think it might make sense to even override `assert_number` for your type! Although a nicer way to do that would be to export an `isnumber` trait.

shashi:
I'm open to PRs for that! Or maybe I can get to it myself soon! I made this change because it's wrong to allow some of these methods on `<:Symbolic{<:Any}`

david-pl:
> Would it make sense in your case to have `struct QNumber <: Number end`? 

Not really. I think I could get it to work, but since `Base` methods are generic for `Number` types it would allow some methods I don't want or lead to some unexpected behavior. Regardless, I suppose someone will eventually need the methods with `Number` for a symtype that is not a `Number` (e.g. symbolic matrices).

> Although a nicer way to do that would be to export an `isnumber` trait.

I think I don't understand what you mean here. I thought `isnumber` is used to check literal number types, i.e. `isnumber(x)==false` and `isnumber(2)==true`. Also, `assert_number` uses `symtype(a)<:Number`. Could you explain how I can use `isnumber` so that `assert_number` passes for my type?

The most convenient way I can think of is to add `assert_number` to the interface (which basically just means documenting it). Maybe also give it a slightly different name, such as `assert_number_methods`? Then users could specify whether a custom symtype should be usable with `Number` by adding a method for their type. What do you think?

shashi:
Oh I didn't mean the `isnumber` function that we already have, we should probably rename that. Okay let me look into making a PR for this!

david-pl:
@shashi Cool, thanks for the quick fix! FYI, the error message mentions a function (`isnumberlike`) that doesn't exist:

ERROR: ArgumentError: The function * cannot be applied to x which is not a Number-like object.Define `isnumberlike(::SymbolicUtils.Sym{QNumber}) = true` to enable this.

PR #136, opened 2020-11-03
Extensible trait-like dispatch for Number methods

shashi:
Solves #134 

cc @YingboMa what do you think?


shashi:
@david-pl You'd have to define `islike(::Symbolic{<:QNumber}, ::Type{Number}) = true` now.

AshtonSBradley:
is there any relation between this and Complex number support for ModelingToolkit?

shashi:
Not right now, but potentially!

ChrisRackauckas:
It's definitely moving along in the right direction.

PR #138, opened 2020-11-19
Removing DataStructures 0.17 from Project.toml

anandijain:
Installing SymbolicUtils caused a good deal of version conflicts for me, I believe this is because of DataStructures 0.17

All tests passed on my fork, so I think they will pass on this PR.

Thanks!


MasonProtter (approved):
CI Failures appear to be unrelated. LGTM.

anandijain:
unrelated but i meant to ask: the add and rm of the Manifest (while not huge) is junk in the tree

is there something that can be done to merge the commits? is this non-issue?

thanks

MasonProtter:
Yeah, generally, people shouldn't include the manifest in a package without a very good reason to include it (needing exact reproducibility). I think this is a benign enough change that it's okay for me to merge without waiting, so I'll do that but if @shashi disapproves we can undo it in the morning.

shashi:
Looks good, thanks! Usually CompatHelper should get this.

PR #139, opened 2020-11-19
fix error message in assertlike

### LINKED ISSUE #134: Non-number symtypes on v0.6

MasonProtter:
In https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/134#issuecomment-721700399, it was pointed out that the error message points to `isnumberlike` instead of `islike`. Closes #134


shashi:
Thanks!

Issue #133, opened 2020-10-23
Division


shashi:
- [ ] make sure inv does not get introduced, only negative powers do.
- [X] simplify using exponent arithmetic
- [ ] put back / after simplification so that there are no negative powers. OR:
- [ ] have an optional argument which simplifies to a rational polynomial normal form i.e. (p / q) where p and q are "polynomials" in variables and non-linear functions of them.

MasonProtter:
Currently, 
```
julia> @syms x
(x,)

julia> x / 37 |> simplify
0.02702702702702703 * x
```
I am starting to wonder if perhaps instead we should just carry around the ` / 37` rather than actually computing `1/37`. It'd be more in line with other CAS and often gives outputs that are more meaningful. Sometimes the `1/37` is rather important. 

People can of course use rationals if they wish, but I wonder if it'd be better for us to symbolically handle rationals for them. Thoughts?

shashi:
I agree, integer division should lead to `term(Rational, ...)`

MasonProtter:
I'm actually not sure if it should lead to `term(Rational, ...)`, because we don't want someone to use `x/2` in their MTK code and then run `build_function` and end up with rationals where they expected fast floating point arithmetic.

MasonProtter:
I think we should just treat `/` with `Int` like a rational, but not actually expose that to the outside code.

shashi:
That works too.

PR #141, opened 2020-11-19
Convert `inv` to literal pow

MasonProtter:
Should accomplish the goal "make sure inv does not get introduced, only negative powers do." in https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/133

I also made a random `global` I noticed a `global const`. Does `MPOLY_CLEANUP` even need to be a `global`, or can it just be internal?

### DIFF
--- a/src/abstractalgebra.jl
+++ b/src/abstractalgebra.jl
@@ -65,10 +65,8 @@ let
                    @acrule(~x::ismpoly * ~y::ismpoly => ~x * ~y)
                    @rule(*(~x) => ~x)
                    @rule((~x::ismpoly)^(~a::isnonnegint) => (~x)^(~a))]
-    MPOLY_CLEANUP = Fixpoint(Postwalk(PassThrough(RestartedChain(mpoly_preprocess))))
+    global const MPOLY_CLEANUP = Fixpoint(Postwalk(PassThrough(RestartedChain(mpoly_preprocess))))
     MPOLY_MAKER = Fixpoint(Postwalk(PassThrough(RestartedChain(mpoly_rules))))
-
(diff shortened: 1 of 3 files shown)


shashi:
Okay I think this could work... Although we should run reverse CI with ModelingToolkit. Because this is the weird part about not having invs:

```julia
julia> g(f)=f(2,-2)
g (generic function with 1 method)

julia> g(^)
ERROR: DomainError with -2:
Cannot raise an integer x to a negative power -2.
Make x or -2 a float by adding a zero decimal (e.g., 2.0^-2 or 2^-2.0 instead of 2^-2), or write 1/x^2, float(x)^-2, x^float(-2) or (x//1)^-2
Stacktrace:
 [1] ^(::Int64, ::Int64) at ./intfuncs.jl:273
 [2] g(::typeof(^)) at ./REPL[19]:1
 [3] top-level scope at REPL[20]:1
 [4] run_repl(::REPL.AbstractREPL, ::Any) at /build/julia/src/julia-1.5.2/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:288

```

`inv` converts the base into float first.

shashi:
When I recently looked into this, `inv`s were actually gone without this fix... I think it could have been because we made things `Real` by default (in MTK at least -- which is what I was testing, but I'm not sure how that should change anything)

MasonProtter:
`inv` was mostly gone, but appears through `Base.literal_pow` before this PR. I'm surprised CI passed, it failed for me locally because of a problem with literal_pow in the fuzzer. I'll make a new PR with the fix.

PR #153, opened 2021-01-04
Update methods.jl

mattwhiteside:
I'm new to the library so take this with a grain of salt, but it seems like this change fixes a problem.  Or at least I found the behavior counter intuitive.

Here's an example:
```julia
julia> @syms x::Real y::Real
(x, y)

julia> typeof(atan(y,x))
SymbolicUtils.Term{Real}

julia> typeof(sqrt(x))
SymbolicUtils.Term{Number}
```

In other words, it looks like diadic functions are maintaining the type of symbols, but monadic functions are promoting a Real to a Number, which could be for example a Complex number, which is not what I was wanting.

Thanks for your work on this library,
-Matt

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -84,7 +84,7 @@ for f in monadic
     if f in [real]
         continue
     end
-    @eval promote_symtype(::$(typeof(f)), T::Type{<:Number}) = Number
+    @eval promote_symtype(::$(typeof(f)), T::Type{<:Number}) = T
     @eval (::$(typeof(f)))(a::Symbolic)   = term($f, a)
 end
 


YingboMa:
We haven't thought about type promotion too seriously yet. For instance, there's
```julia
julia> x |> typeof
Sym{Int64}

julia> hypot(x, x) |> typeof
Term{Int64}
```
We need to have a way to handle these cases more systematically.

mattwhiteside:
Thanks for taking a look.  Your comment makes sense.

shashi:
This PR has good intention, but it is wrong for, say, `sin(::Int)` because the output is not Int.

mattwhiteside:
Ok yes, I see your point.

Maybe another way of stating my confusion... it seems straightforward and reasonable to promote an `::Int` to a `::Real,` but promoting a `::Real` to a `::Number`, which encompasses complex numbers, seems like a less straightforward promotion, due to the 2-dimensional aspect of complex numbers.

mattwhiteside:
Hi Karl, 

I made some notes in the attached screenshot.

I finished chapter 4 and 5 of Barbour (chapter 5 seemed shorter).   I keep having to look back and remind myself what the ’Shuetz-Boltzmann fluctuation hypothesis’ is, but I think I’m roughly following his argument: that configurations of low entropy (aka high organization or complexity), should be short and fleeting, and that entropy should increase both coming into and out of this short high entropy state, but what we observe is that the entropy only increases moving away from this state, aka forward in time.

-Matt

shashi:
@mattwhiteside seems like you replied to the wrong email thread :)

mattwhiteside:
Deleted.  Thanks.

PR #158, opened 2021-01-09
[WIP] adding a few rules for trigonometric functions

crstnbr:
In this PR I'm trying to add double-angle and addition/subtraction formulas, see for example https://mathworld.wolfram.com/TrigonometricAdditionFormulas.html.

I added the corresponding rules but currently this doesn't seem to have any effect. What am I missing?

Once I figured out how to make this work I will add a few tests.

### DIFF
--- a/src/simplify_rules.jl
+++ b/src/simplify_rules.jl
@@ -66,7 +66,24 @@ let
         @acrule(cot(~x)^2 + -1*csc(~x)^2 => one(~x))
         @acrule(cot(~x)^2 +  1 => csc(~x)^2)
         @acrule(csc(~x)^2 + -1 => cot(~x)^2)
-    ]
+
+        # double-angle rules
+        @acrule(cos(~x)^2 - sin(~x)^2   => cos(2 * ~x))
+        @acrule(2*sin(~x)cos(~x)        => sin(2 * ~x))
+        @acrule(cosh(~x)^2 + sinh(~x)^2 => cosh(2 * ~x))
+        @acrule(2*sinh(~x)cosh(~x)      => sinh(2 * ~x))
+
+        # addition/subtraction formulas
+        @acrule(sin(~x)cos(~y) + cos(~x)sin(~y) => sin(~x + ~y))
+        @acrule(sin(~x)cos(~y) - cos(~x)sin(~y) => sin(~x - ~y))
+        @acrule(cos(~x)cos(~y) + sin(~x)sin(~y) => cos(~x - ~y))
+        @acrule(cos(~x)cos(~y) - sin(~x)sin(~y) => cos(~x + ~y))
+        
+        @acrule(sinh(~x)cosh(~y) + cosh(~x)sinh(~y) => sinh(~x + ~y))
+        @acrule(cosh(~x)cosh(~y) + sinh(~x)sinh(~y) => cosh(~x + ~y))
+        @acrule(sinh(~x)cosh(~y) - cosh(~x)sinh(~y) => sinh(~x - ~y))
+        @acrule(cosh(~x)cosh(~y) - sinh(~x)sinh(~y) => cosh(~x - ~y))
+        ]
 
     BOOLEAN_RULES = [
         @rule((true | (~x)) => true)


shashi:
```julia
julia> ar = @acrule(sin(~x)cos(~y) + cos(~x)sin(~y) => sin(~x + ~y))
ACRule(sin(~x) * cos(~y) + cos(~x) * sin(~y) => sin(~x + ~y))

julia> ar(sin(a)cos(b) + cos(a)sin(b))
sin(a + b)
```
this works

so it should work...

> I added the corresponding rules but currently this doesn't seem to have any effect.

How do you mean?

src/simplify_rules.jl
```diff
@@ -66,7 +66,24 @@ let
         @acrule(cot(~x)^2 + -1*csc(~x)^2 => one(~x))
         @acrule(cot(~x)^2 +  1 => csc(~x)^2)
         @acrule(csc(~x)^2 + -1 => cot(~x)^2)
-    ]
+
+        # double-angle rules
+        @acrule(cos(~x)^2 - sin(~x)^2   => cos(2 * ~x))
```
MasonProtter, line 71:
```suggestion
        @acrule(cos(~x)^2 + -1*sin(~x)^2   => cos(2 * ~x))
```
this is important not only because of how we represent `x - y` as `x + -1y`, but also because you're using `acrule` here, which stands for 'associative commutative', but subtraction is not ac.

src/simplify_rules.jl
```diff
@@ -66,7 +66,24 @@ let
...
+        @acrule(2*sin(~x)cos(~x)        => sin(2 * ~x))
```
MasonProtter, line 72:
```suggestion
        @acrule(2*sin(~x)*cos(~x)        => sin(2 * ~x))
```
Explicit `*` seem to be needed here to get the simplification to happen.

src/simplify_rules.jl
```diff
@@ -66,7 +66,24 @@ let
...
+        @acrule(cosh(~x)^2 + sinh(~x)^2 => cosh(2 * ~x))
+        @acrule(2*sinh(~x)cosh(~x)      => sinh(2 * ~x))
```
MasonProtter, line 74:
```suggestion
        @acrule(2*sinh(~x)*cosh(~x)      => sinh(2 * ~x))
```

src/simplify_rules.jl
```diff
@@ -66,7 +66,24 @@ let
...
+
+        # addition/subtraction formulas
+        @acrule(sin(~x)cos(~y) + cos(~x)sin(~y) => sin(~x + ~y))
+        @acrule(sin(~x)cos(~y) - cos(~x)sin(~y) => sin(~x - ~y))
```
MasonProtter, line 78:
```suggestion
        @acrule(sin(~x)cos(~y) + -1*cos(~x)*sin(~y) => sin(~x - ~y))
```

src/simplify_rules.jl
```diff
@@ -66,7 +66,24 @@ let
...
+        @acrule(cos(~x)cos(~y) + sin(~x)sin(~y) => cos(~x - ~y))
+        @acrule(cos(~x)cos(~y) - sin(~x)sin(~y) => cos(~x + ~y))
```
MasonProtter, line 80:
```suggestion
        @acrule(cos(~x)cos(~y) + -1*sin(~x)*sin(~y) => cos(~x + ~y))
```

src/simplify_rules.jl
```diff
@@ -66,7 +66,24 @@ let
...
+        
+        @acrule(sinh(~x)cosh(~y) + cosh(~x)sinh(~y) => sinh(~x + ~y))
+        @acrule(cosh(~x)cosh(~y) + sinh(~x)sinh(~y) => cosh(~x + ~y))
+        @acrule(sinh(~x)cosh(~y) - cosh(~x)sinh(~y) => sinh(~x - ~y))
```
MasonProtter, line 84:
```suggestion
        @acrule(sinh(~x)cosh(~y) + -1*cosh(~x)*sinh(~y) => sinh(~x - ~y))
```

src/simplify_rules.jl
```diff
@@ -66,7 +66,24 @@ let
...
+        @acrule(cosh(~x)cosh(~y) - sinh(~x)sinh(~y) => cosh(~x - ~y))
```
MasonProtter, line 85:
```suggestion
        @acrule(cosh(~x)cosh(~y) + -1*sinh(~x)*sinh(~y) => cosh(~x - ~y))
```

MasonProtter (changes requested):
I think with these tweaks and with appropriate tests, this should be good to go.

crstnbr:
I'm in the process of writing a few tests.  I noticed that while the rules work in the first case they don't in the second. Why?

```Julia
julia> simplify((cos(x)^2 - sin(x)^2)) - cos(2x)
0

julia> simplify((cos(x)^2 - sin(x)^2) - cos(2x)) # Why don't we get 0 here as well?
(cos(x)^2) + -1(cos(2x) + (sin(x)^2))
```

crstnbr:
Another thing that I don't understand.

```Julia
julia> ar = @acrule(2*sinh(~x)*cosh(~x)      => sinh(2 * ~x))
ACRule(2 * sinh(~x) * cosh(~x) => sinh(2 * ~x))

julia> ar(2*sinh(x)*cosh(x))
sinh(2x)
```

but

```Julia
julia> @syms x::Real
(x,)

julia> simplify(2*sinh(x)*cosh(x))
2cosh(x)*sinh(x)
```

crstnbr:
If someone could give me some feedback on why the examples above don't work as expected I would try to finish this PR soon.

garrekds:
I ran into issues with trig identities not working as expected when doing matrix multiplication, specifically when multiplying a matrix by its inverse. I'll copy relevant portions from a post I made on discourse.

I also tested the trig identities that @crstnbr wrote and they fail in the same way on my computer. I copied his tests into rulesets.jl and the same ones passed/failed. It seems like some of the rules work and others don't or the rules don't kick in when the expression is slightly different.


Here is an example of the issue I found using the 2x2 rotation matrix:
```
julia> using LinearAlgebra, Symbolics, SymbolicUtils

julia> @variables x::Real;
julia> I2 = [1 0; 0 1];     #identity matrix

julia> A = [cos(x) -sin(x); sin(x) cos(x)]
2×2 Matrix{Num}:
 cos(x)  -sin(x)
 sin(x)   cos(x)

julia> simplify.(inv(A) * A)
2×2 Matrix{Num}:
 1  …                                                   sin(x)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1) - (sin(x)*(true - ((cos(x)^-1)*(sin(x)^2)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1)))*(cos(x)^-1))
 0     cos(x)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1) + (cos(x)^-1)*(sin(x)^2)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1)
```
By eye, the result should simplify to I2 (except for those true values). I think the issue has something to do with the LU factorization going on. Next I tried directly accessing and using left-division instead of `inv` (i.e. solving A^-1 = A / I):

```
julia> simplify.(lu(A) \ I2 * A)
2×2 Matrix{Num}:
 1  …  sin(x)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1) - (sin(x)*(1 - ((cos(x)^-1)*(sin(x)^2)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1)))*(cos(x)^-1))
 0                                  cos(x)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1) + (cos(x)^-1)*(sin(x)^2)*((cos(x) + (cos(x)^-1)*(sin(x)^2))^-1)
```

Again, it evaluates the expression correctly, but can’t simplify further (even though it should be able to).
When setting x = π/4 (or any other value) using substitute, the expression correctly reduces to the identity matrix, but some of the elements are converted to floats:

```
julia> B = inv(A) * A;

julia> answer = simplify.(substitute.(B, (Dict(x => π/4),)))
2×2 Matrix{Num}:
 1  0.0
 0  1.0

julia> answer == I2
true
```
I found one workaround, bypassing `inv` altogether, but it only works for 2x2 matrices.

Using the Caley-Hamilton method ([Wikipedia](https://en.wikipedia.org/wiki/Invertible_matrix#Analytic_solution)):
```
julia> A_inv = det(A) * (tr(A) * I2 - A)
2×2 Matrix{Num}:
  cos(x)*(cos(x)^2 + sin(x)^2)  sin(x)*(cos(x)^2 + sin(x)^2)
 -sin(x)*(cos(x)^2 + sin(x)^2)  cos(x)*(cos(x)^2 + sin(x)^2)

julia> simplify.(A_inv * A)
2×2 Matrix{Num}:
 1  0
 0  1

julia> simplify.(A_inv * A) == I2
true
```

yewalenikhil65:
@carstenbauer  is there plan to include product to sum rule similar to TrigReduce in wolfram through this PR? https://reference.wolfram.com/language/ref/TrigReduce.html

carstenbauer:
Note that I haven't looked at this PR in two years (and also don't plan to move on with it any time soon). Feel free to take it over!

PR #161, opened 2021-01-12
Ensure transitivity in tree ordering

### LINKED ISSUE #160: <ₑ is not transitive
With
```julia
using Latexify
using ModelingToolkit

@parameters t σ ρ β
@variables x(t) y(t) z(t)
@derivatives D'~t

eqs = [D(x) ~ σ*(y-x)*D(x-y)/D(z),
       0 ~ σ*x*(ρ-z)/10-y,
       D(z) ~ x*y^(2//3) - β*z]
```
we have
```julia
julia> using SymbolicUtils: arguments, operation, <ₑ

julia> args = arguments(eqs[1].rhs)
4-element Array{SymbolicUtils.Symbolic,1}:
 σ
 (D'~t)(x(t) + -1y(t))
 y(t) + -1x(t)
 (D'~t)(z(t))^-1

julia> args[1] <ₑ args[2] <ₑ args[3] <ₑ args[4]
true

julia> args[1] <ₑ args[4]
false
```

YingboMa:
Fix #160

### DIFF
--- a/src/ordering.jl
+++ b/src/ordering.jl
@@ -13,39 +13,13 @@ function <ₑ(a, b)
     if !istree(a) && !istree(b)
         T = typeof(a)
         S = typeof(b)
-        T===S ? isless(a, b) : nameof(T) < nameof(S)
+        return T===S ? isless(a, b) : nameof(T) < nameof(S)
     elseif istree(b) && !istree(a)
-        args = arguments(b)
-        if length(args) === 2
-            n1, n2 = !is_literal_number(args[1]) , !is_literal_number(args[2])
-            if n1 && n2
-                # both subterms are terms, so it's definitely firster
-                return true
-            elseif n1
-                return isequal(a, args[1]) || a <ₑ args[1]
-            elseif n2
-                return isequal(a, args[2]) || a <ₑ args[2]
-            else
-                # both arguments are not numbers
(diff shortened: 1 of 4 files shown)


src/ordering.jl
```diff
@@ -13,39 +13,13 @@ function <ₑ(a, b)
...
-            if isequal(a, args[1])
-                return true # e.g sin(a)*a should become a*sin(a)
-            else
-                return a<ₑargs[1]
-            end
-        else
-            # variables to the right
-            return false
-        end
+        return true
```
shashi, line 18: Lol!! This was there so that you can get `a*sin(a)*b*b^2*sin(b)` I tried to more or less follow guide to the perplexed.

YingboMa: It's very computational intensive to do that kind of ordering, since you have to traverse the whole tree to check it. SymEngine doesn't do it.

```julia
julia> using SymEngine

julia> @vars x y z
(x, y, z)

julia> sin(y) + x + x^2 + y^2 + z^2
x + x^2 + y^2 + z^2 + sin(y)
```

shashi: True, but I'd like to keep a slow comparison that prints well. We can write a fast comparison for non-human purposes later.

PR #164, opened 2021-01-12
Fix division and printing

### LINKED ISSUE #163: Bug in division rules
This is in SymbolicUtils v0.7.2:
```julia
julia> using SymbolicUtils

julia> @syms μ σ x
(μ, σ, x)

julia> simplify(-log(σ) - (x - μ)^2 / (2*σ^2))
-1log(σ) + -2(σ^-2)*((x + -1μ)^2)
```

That `-2` should be `-0.5`. Looks like a bug in the rules?

YingboMa:
Fix #163

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -351,7 +351,7 @@ function show_term(io::IO, t)
 
                 if i != length(args)
                     if fname == :*
-                        if i == 1 && args[1] isa Number && !paren_scalar
+                        if i == 1 && args[1] isa Number && !(args[2] isa Number) && !paren_scalar
                             # skip
                             # do not show * if it's a scalar times something
                         else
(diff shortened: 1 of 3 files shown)


src/types.jl
```diff
@@ -648,8 +645,9 @@ Base.show(io::IO, p::Pow) = show_term(io, p)
 ^(a::Number, b::SN) = Pow(a, b)
 
 function ^(a::Mul, b::Number)
+    coeff = a.coeff isa Integer && b isa Integer ? (a.coeff//1) ^ b : a.coeff ^ b
```
shashi, line 648: Cool trick!

shashi (approved):
Lgtm!

PR #165, opened 2021-01-14
Terms constructed by substitution should be the same as terms constructed by evaluation

YingboMa:
We need `symtype(similarterm(f, args...)) === symtype(f(args...))` to
make sure that terms constructed by substitution are the same as terms
constructed by evaluation.

This PR also fixes `isequal(::Add, ::Symbolic)`-like checks.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -134,8 +134,12 @@ for (f, Domain) in [(==) => Number, (!=) => Number,
     end
 end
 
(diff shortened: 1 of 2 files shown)


shashi:
Since we did this I think we need to provide a way of forcing the type to be something known.

This means we need to change the `similarterm` interface. This would be the correct way to fix #175 and #193

PR #168, opened 2021-01-15
Fix `rec_promote_symtype` so that it only recurse on VARARGFUNS

YingboMa:
Fix #167

PR #169, opened 2021-01-15
use mapfoldl instead of rec_promote_symtype, fix it on Sym{<:FnType}

shashi:
supersedes #168


shashi:
I believe this qualifies as a patch release even though it unbroke a broken API? :)

PR #172, opened 2021-01-16
Hash cache with 0 salt

shashi:
caches in more cases than #171

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -296,10 +296,11 @@ arguments(x::Term) = getfield(x, :arguments)
 hashvec(xs, z) = foldr(hash, xs, init=z)
 
 function Base.hash(t::Term{T}, salt::UInt) where {T}
+    !iszero(salt) && hash(hash(t, zero(UInt)), salt)
     h = t.hash[]
-    (iszero(salt) && !iszero(h)) && return h
+    !iszero(h) && return h
     h′ = hashvec(arguments(t), hash(operation(t), hash(T, salt)))
-    iszero(salt) && (t.hash[] = h′)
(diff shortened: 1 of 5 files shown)


src/types.jl
```diff
@@ -296,10 +296,11 @@ arguments(x::Term) = getfield(x, :arguments)
 hashvec(xs, z) = foldr(hash, xs, init=z)
 
 function Base.hash(t::Term{T}, salt::UInt) where {T}
+    !iszero(salt) && hash(hash(t, zero(UInt)), salt)
```
YingboMa, line 299: Shouldn't there be a return?

YingboMa:
```suggestion
    !iszero(salt) && return hash(hash(t, zero(UInt)), salt)
```

shashi: yup

src/types.jl
```diff
@@ -296,10 +296,11 @@ arguments(x::Term) = getfield(x, :arguments)
...
     h = t.hash[]
-    (iszero(salt) && !iszero(h)) && return h
+    !iszero(h) && return h
     h′ = hashvec(arguments(t), hash(operation(t), hash(T, salt)))
-    iszero(salt) && (t.hash[] = h′)
+    (t.hash[] = h′)
```
YingboMa, line 303:
```suggestion
    t.hash[] = h′
```

shashi:
MTK fails looking weird.

YingboMa:
No... Tests fails with this PR should probably don't merge it.

YingboMa:
It was passing before

PR #176, opened 2021-01-20
fix Add - Add

shashi:


YingboMa:
The bug is that the `f` in `_merge` must be a unary operation.

PR #180, opened 2021-01-22
Enable ordered AC rules for user

MilesCranmer:


shashi:
this needs some documentation :sweat_smile: I thought it's a bit hard to explain, so just left it indoors.

PR #189, opened 2021-01-29
deprecate cond for IfElse.ifelse

oxinabox:
I couldn't find where the tests i should update are.
Let me know where is should add some.

This is what `IfElse.ifelse` is for so lets use that


src/methods.jl
```diff
@@ -137,11 +137,11 @@ end
...
 
 # An ifelse node, ifelse is a built-in unfortunately
-#
-cond(_if::Bool, _then, _else) = ifelse(_if, _then, _else)
-function cond(_if::Symbolic{Bool}, _then, _else)
+# So this uses IfElse.jl's ifelse that we imported
+function ifelse(_if::Symbolic{Bool}, _then, _else)
     Term{Union{symtype(_then), symtype(_else)}}(cond, Any[_if, _then, _else])
 end
+Base.@deprecate cond(_if, _then, _else) ifelse(_if, _then, _else)
```
oxinabox, line 144: The deprecation warning this prints is mildly misleading.
It prints
```julia
┌ Warning: `cond(_if, _then, _else)` is deprecated, use `ifelse(_if, _then, _else)` instead.
```
Because I can't write `IfElse.ifelse` or it will name clash with the rewriter.

Maybe i should hand implement the deprecation warning to force it to say `IfElse.ifelse`

shashi:
I hadn't done this because we export a type called `IfElse` from `Rewriters` sub-module.... I would be fine with renaming that to `RIfElse` with a deprecation. It should be possible in theory. If a package does `using SymbolicUtils.Rewriters`, it cannot do `using IfElse`...

oxinabox:
This is true. But.
Don't need to do `using IfElse`.
Can just do `import IfElse: ifelse` or `using IfElse: ifelse`

Plus in 1.6 can do `inport IfElse as IE`

src/methods.jl
shashi: I think this is fine... But you're welcome to do the manual thing.

Issue #196, opened 2021-02-03
symtype(x) = Any


cscherrer:
Hi,

I think this fallback may be causing trouble, because any concrete value in a symbolic expression will cause the whole thing to be `::Any`. But in these cases we know the _actual_ type. Is there a reason for discarding this?

What do you think of changing this to
```julia
symtype(x) = typeof(x)
```
?

YingboMa:
Yeah, that seems to be more sensible. Could you make a PR?

PR #197, opened 2021-02-03
default to symtype(x) = typeof(x)

cscherrer:
This tiny PR just changes

```julia
symtype(x) = Any
```

to

```julia
symtype(x) = typeof(x)
```

The current implementation leads to a loss of type information for many non-symbolic values. This propagates up the tree, leading to lots of expressions erroneously being considered `::Any`

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -52,7 +52,7 @@ function symtype end
 
 symtype(x::Number) = typeof(x)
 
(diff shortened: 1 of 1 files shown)


cscherrer:
This PR fixes https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/196

PR #198, opened 2021-02-03
Allow StaticArrays 0.12

cscherrer:
Soss has some dependencies that haven't yet updated their `StaticArrays` compat to 1.0. Could we loosen this a little?


ChrisRackauckas (approved):
definitely no reason to require 1.0

Issue #121, opened 2020-08-15
Common subexpression elimination


shashi:
The basics:

```julia
using SymbolicUtils
using SymbolicUtils: Sym, Term
using SymbolicUtils.Rewriters
using DataStructures

newsym() = Sym{Number}(gensym("cse"))

function cse(expr)
    dict = OrderedDict()
    r = @rule ~x::(x -> x isa Term) => haskey(dict, ~x) ? dict[~x] : dict[~x] = gensym()
    final = Postwalk(Chain([r]))(expr)
    [[var=>ex for (ex, var) in pairs(dict)]..., final]
end
```
Examples:
```julia
@syms x y
julia> cse(cos(cos(x)) + sin(cos(x)))
5-element Array{Any,1}:
 Symbol("##260") => cos(x)
 Symbol("##261") => cos(##260)
 Symbol("##262") => sin(##260)
 Symbol("##263") => ##261 + ##262
                  Symbol("##263")

julia> SymbolicUtils.show_simplified[] = false
false

julia> cse(cos(cos(x)) + sin(cos(x)))
5-element Array{Any,1}:
 Symbol("##264") => cos(x)
 Symbol("##265") => cos(##264)
 Symbol("##266") => sin(##264)
 Symbol("##267") => ##265 + ##266
                  Symbol("##267")

julia> cse(cos(cos(x)) + cos(cos(x)))
4-element Array{Any,1}:
 Symbol("##268") => cos(x)
 Symbol("##269") => cos(##268)
 Symbol("##270") => ##269 + ##269
                  Symbol("##270")
```

This issue is a good place to think about some API questions:

1. What should CSE return
2. Should we be able to place that thing in a different Term?

cscherrer:
Hi @shashi , I had modified this to work in my context, but it's no longer working. Could something have changed in the last merge that broke it?

My setup is like this:
```julia
function cse(s::Symbolic)
    vars = atoms(s)
    dict = OrderedDict()
    r = @rule ~x => csestep(~x, vars, dict) 
    final = RW.Postwalk(RW.PassThrough(r))(s)
    [[var=>ex for (ex, var) in pairs(dict)]...]
end

export csestep

csestep(s::Sym, vars, dict) = s

csestep(s, vars, dict) = s

function csestep(x::S, vars, dict) where {S <: Symbolic}
    # Avoid breaking local variables out of their scope
    isempty(setdiff(atoms(x), vars)) || return x

[...]

PR #200, opened 2021-02-03
WIP: CSE

shashi:
trying out a solution to #121 with `Let`

### DIFF
--- a/src/code.jl
+++ b/src/code.jl
@@ -469,4 +469,58 @@ function toexpr(exp::LiteralExpr, st)
     recurse_expr(exp.ex, st)
 end
 
+
+### Code-related utilities
+
+using SymbolicUtils
+
+using SymbolicUtils: Sym, Term
+
+using SymbolicUtils.Rewriters
+
+using DataStructures
+
+### Common subexprssion evaluation
+
+newsym() = Sym{Number}(gensym("cse"))
+
+function _cse(expr, dict=OrderedDict())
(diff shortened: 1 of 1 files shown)


cscherrer:
In case it's helpful, I have a full implementation here:
https://github.com/cscherrer/SymbolicCodegen.jl/blob/main/src/cse.jl

dpad:
@shashi Thanks for your work on this! I've managed to use your WIP work so far (plus some additional functions for my use case) here https://github.com/dpad/OrbitalTrajectories.jl/pull/7 to regain the performance improvements from https://github.com/SciML/ModelingToolkit.jl/pull/698.

dpsanders:
I need an option to split up terms like `3x^2 * y` into `u = x^2; v = 3*u; w = v * y`.
I am trying to see how to do this but it's pretty unclear to me.

dpsanders:
I now have a working, but pedestrian version of this (not using `@rule`) here: 

https://github.com/dpsanders/ReversePropagation.jl/blob/a5f3d9fe86e76ca7272faadeca6f8dd274c0b6cb/src/cse_new.jl

YingboMa:
I think we should just do it via a list of assignments as oppose to `Let`, since the printed expression won't be readable at all if we were to use let.

YingboMa:
@shashi I optimized the CSE function a bit. Here's the result:
```julia
julia> @btime cse($(Nu_sym.val)); # before
  930.986 μs (6258 allocations: 249.12 KiB)

julia> @btime cse($(Nu_sym.val)); # after
  680.307 μs (4600 allocations: 204.91 KiB)
```

PR #204, opened 2021-02-16
Par for parallelism

shashi:
Maybe a better name is `SpawnFetch`?


shashi:
another place which uses Par https://wiki.haskell.org/Par_and_seq

PR #215, opened 2021-02-28
WIP: Metadata system take 2

shashi:
Now always keyed by a context type.


src/types.jl
```diff
@@ -56,6 +56,55 @@ symtype(x) = typeof(x)
...
+    end
+    Base.ImmutableDict{DataType, Any}(d, ctx, val)
+end
+
+function setmetadata(s::Symbolic, ctx::DataType, val)
+    if s.metadata isa Ref
+        @set s.metadata = Ref{Any}(assoc(s.metadata[], ctx, val))
+    else
+        # fresh Dict
+        @set s.metadata = Ref{Any}(Base.ImmutableDict{DataType, Any}(ctx, val))
```
shashi, line 104: @YingboMa I think this Ref is not necessary here, it will be `Union{Nothing, ImmutableDict{DataType, Any}}`.

shashi: Just as good. In fact we can even use `ImmutableDict{DataType, Any}()` as the default "empty" thing, which would mean no type parameter. But that empty struct adds 24 bytes to Sym... Which maybe just fine!

YingboMa: Have you benchmarked it?

shashi: No I didn't benchmark the empty ImmutableDict, right now it is still a parameter which can be Nothing or ImmutableDict.

shashi: This is faster than having a Ref of ImmutableDict.

PR #219, opened 2021-03-02
Make metadata a keyword

YingboMa:
Positional arguments are used for indices in Symbolics.jl

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -175,7 +175,7 @@ Base.nameof(s::Sym) = s.name
 
 ConstructionBase.constructorof(s::Type{<:Sym{T}}) where {T} = Sym{T}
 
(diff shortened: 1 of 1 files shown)


shashi:
Oops, okay this needs some fixes.

shashi:
The reverse CI didn't catch this :roll_eyes:

ChrisRackauckas:
You know what that means. https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/220

shashi:
then we gotta remove this https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/master/test/runtests.jl#L25-L30

shashi:
DDDiffEq fails with "WARNING: both ModelingToolkit and DataDrivenDiffEq export "jacobian"; uses of it in module Main must be qualified" unrelated but just pointing it out.

PR #229, opened 2021-03-08
default for `polynorm` should be `true`, fixed

JeffreySarnoff:
I did not bump the version number bcaws I do not know how you are using it, or if more than one change goes into the next step.


JeffreySarnoff:
what -- errorre -- tahw

it is a simple fix -- line 15 in api.jl ` = true`

PR #231, opened 2021-03-09
WIP rational polynomial form

shashi:


sumiya11:
Hi, @shashi 

I would like to clarify a couple of details on the desired implementation for these polynomial fractions.

- Seems to me the current `Div` partly shares interface with structures from src/types.jl, for example, `Mul`. Is that OK for now?
- Each call to `/(a::Union{SN,Number}, b::SN)` creates a new `PolynomialRing`. So, a single symbol may be mapped to many different `MPoly` variables for different `Div` instances. Writing a simple `isequal(::Div, ::Div)` operator might be tricky as a consequence. Perhaps it will be convenient to keep track of all `MPoly` variable creations and use them later?.. If it actually is, could you share thoughts on how to implement it better?

src/polyform.jl
```diff
@@ -0,0 +1,95 @@
...
+function operation(x::PolyForm)
+    MP.nterms(x.p) == 1 ? (*) : (+)
+end
+
+function arguments(x::PolyForm{T}) where {T}
+
+    function is_var(v)
+        MP.nterms(v) == 1 &&
+        isone(sum(x->abs(MP.degree(v, x)), MP.variables(MP.monomial(v))))
+    end
```
shashi, line 71: @blegat I adapted this from `convert(PolyVar, x)` let me know if this is correct. PolyForm is supposed to behave like a Term (in terms of the interface) but use a polynomial as the underlying representation.

shashi:
@YingboMa have a look, the code is much simpler than the stuff in abstractalgebra.jl :)

I am still wondering if we should implement `similarterm` on PolyForm, and if so, what would it look like.

src/polyform.jl
```diff
@@ -0,0 +1,95 @@
...
+    PolyForm{symtype(x)}(polyize(x, pvar2sym, sym2term, vtype),
+                         pvar2sym, sym2term, metadata(x))
+end
+
+PolyForm(x, args...) = x
+
+istree(x::PolyForm) = true
+
+function operation(x::PolyForm)
+    MP.nterms(x.p) == 1 ? (*) : (+)
```
YingboMa, line 63: What is `x.p` is just the polyvar `x`?

YingboMa (commented):
Woah, this is amazingly simple and elegant. Gotta love closures.

src/polyform.jl
shashi: You mean "what if"? All the interface functions seem to work.

src/polyform.jl
```diff
@@ -0,0 +1,199 @@
...
+            name = Symbol(string(op), "-", hash(x))
+
+            @label lookup
+            sym = Sym{symtype(x)}(name)
+            if haskey(sym2term, sym)
+                if isequal(sym2term[sym][1], x)
+                    return sym2term[sym][2]
+                else # hash collision
+                    name = Symbol(name, "_")
+                    @goto lookup
```
shashi, line 56: @YingboMa you might appreciate this hash collision management :rofl:

shashi:
this works now:

```julia
julia> simplify_fractions(((x-y+z)*(x+4z+1)) / (y*(2x - 3y + 3z) + x*(x + z)))
((x + 4z + 1)) / ((x + 3.0y))
```

shashi:
I think we should just use `Mul` instead of a vector of factors for numerator and denominator! Since supporting

```
simplify_fractions((x+y+z-1)^3/(x+y-1))
```

without expanding would require us to maintain the same data structure as Mul anyway....  Mul, Div, Pow should all work by creating each other, maybe.

src/simplify_rules.jl
```diff
@@ -39,7 +39,7 @@ let
         @rule((((~x)^(~p::_isinteger))^(~q::_isinteger)) => (~x)^((~p)*(~q)))
         @rule(^(~x, ~z::_iszero) => 1)
         @rule(^(~x, ~z::_isone) => ~x)
-        @rule(inv(~x) => ~x ^ -1)
+        @rule(inv(~x) => 1/(~x))
```
shashi, line 42: should we even keep this one? @YingboMa

PR #238, opened 2021-03-16
Add `function_to_expr` to make `toexpr` of calls extendible

YingboMa:


shashi:
Should we just use `toexpr`?

src/code.jl
```diff
@@ -96,23 +96,38 @@ Base.convert(::Type{Assignment}, p::Pair) = Assignment(pair[1], pair[2])
...
             return toexpr(Term{Any}(inv, [ex]), st)
         else
             return toexpr(Term{Any}(^, [Term{Any}(inv, [ex]), -args[2]]), st)
         end
-    elseif op === (SymbolicUtils.ifelse)
-        return :($(toexpr(args[1], st)) ? $(toexpr(args[2], st)) : $(toexpr(args[3], st)))
-    elseif op isa Sym && haskey(st.symbolify, O)
-        return st.symbolify[O]
     end
-    return Expr(:call, toexpr(op, st), map(x->toexpr(x, st), args)...)
```
shashi, line 115: This already did `toexpr(op, st)` why is that not sufficient?

YingboMa: Do you mean to move the generic case to `toexpr`, too?

YingboMa:
> Should we just use toexpr?

I thought about this. But I thought it might be harder to write unambiguous methods on `toexpr`.

PR #245, opened 2021-03-24
MakeArray compatible with subarray

tmigot:
If the similar type is a `SubArray`, the `MakeArray` returns a classical `Array`.
This is necessary as it is not a subtype of `Array`.


shashi:
A better implementation would be to use the parent type?

```
@inline create_array(S::Type{<:SubArray}, T, dims, elems...)  = create_array(parent_type(S), T, dims, elems...)
```
I made up `parent_type`, but I'm sure there's something like that.

YingboMa:
Right. We can also handle `SubArray`, `ReshapedArray`, `LinearAlgebra.Transpose`, and `LinearAlgebra.Adjoint`.

tmigot:
Thanks, both for the feedback. @shashi you are right we can do something similar, using parametric types of `SubArray{T,N,P,I,L}`.
I made a shortlist for such arrays: in Base.jl `SubArray`, `PermutedDimsArray`, `ReinterpretedArray`, `ReshapedArray`; and in LinearAlgebra.jl `Adjoint`, `Transpose`, `Symmetric`, `Hermitian`.

All these types have a `parent` function, but not a `parent_type` function (yet?). So, I will make a function for each.

shashi:
Thorough! :+1:

shashi:
Very neat! :tada:

PR #246, opened 2021-03-24
Make print_tree show coefficients for Add and Mul

adamslc:
Also makes printing less verbose when the coefficient/power on a given term is one.
Before:
```
3 + x + 5y
+
├─ (*)
│  ├─ x
│  └─ 1
└─ (*)
   ├─ y
   └─ 5

2x*(y^-1)
*
├─ (^)
│  ├─ x
│  └─ 1
└─ (^)
   ├─ y
   └─ -1
```
After:
```
3 + x + 5y
+
├─ 3
├─ x
└─ (*)
   ├─ y
   └─ 5

2x*(y^-1)
*
├─ 2
├─ x
└─ (^)
   ├─ y
   └─ -1
```

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -902,7 +902,17 @@ struct TreePrint
     x
 end
 AbstractTrees.children(x::Term) = arguments(x)
-AbstractTrees.children(x::Union{Add, Mul}) = map(y->TreePrint(x isa Add ? (:*) : (:^), y), collect(pairs(x.dict)))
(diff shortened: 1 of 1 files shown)


YingboMa:
Thanks for the PR! If memory serves, we actually intensionally print Add and Mul like that to differentiate them from other terms. But I think this printing is fine, too. @shashi what do you think?

shashi:
This is fine since there’s no ambiguity.

PR #254, opened 2021-03-28
Specialize `Base.isequal`

YingboMa:
Fix https://github.com/JuliaSymbolics/Symbolics.jl/issues/129

```julia

using MeshCat
using ModelingToolkit
using TORA
using RigidBodyDynamics
using Symbolics
import SymPy

function setup(vis, n_dof, sympy=false)
    robot = create_robot_kuka_iiwa_14(vis)
    x = MechanismState{sympy ? SymPy.Sym : Num}(robot.mechanism)
    q = configuration(x)
    if sympy
        for i in 1:n_dof
            q[i] = SymPy.symbols("q_$i", real = true)
        end
    else
        for i in 1:n_dof
            q[i] = Num(Sym{Real}(Symbol("q$i")))
        end
    end
    return x
end
vis = Visualizer()
```

Symbolics:
```julia
julia> x = setup(vis, 3); @time M = mass_matrix(x);
  3.721925 seconds (7.73 M allocations: 462.482 MiB, 2.41% gc time, 99.20% compilation time)

julia> for n in 3:7; x = setup(vis, n); @time M = mass_matrix(x); end
  0.006976 seconds (79.61 k allocations: 3.028 MiB)
  0.007024 seconds (80.71 k allocations: 3.072 MiB)
  0.007367 seconds (81.56 k allocations: 3.106 MiB)
  0.007341 seconds (82.07 k allocations: 3.131 MiB)
  0.007315 seconds (82.66 k allocations: 3.157 MiB)
```

SymPy:
```julia
julia> x = setup(vis, 3, true); @time M = mass_matrix(x);
 16.333147 seconds (31.71 M allocations: 1.870 GiB, 2.18% gc time)

julia> for n in 3:7; x = setup(vis, n, true); @time M = mass_matrix(x); end
  2.899745 seconds (78.61 k allocations: 2.414 MiB)
  4.737822 seconds (78.61 k allocations: 2.414 MiB)
  9.462235 seconds (78.61 k allocations: 2.414 MiB)
 16.858959 seconds (78.61 k allocations: 2.414 MiB)
 17.311054 seconds (78.61 k allocations: 2.414 MiB)
```

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -102,12 +102,19 @@ function setmetadata(s::Symbolic, ctx::DataType, val)
     end
 end
 
-Base.isequal(s::Symbolic, x) = false
-Base.isequal(x, s::Symbolic) = false
+Base.isequal(::Symbolic, x) = false
+Base.isequal(x, ::Symbolic) = false
+Base.isequal(::Symbolic, ::Symbolic) = false
 
-function Base.isequal(t1::Symbolic, t2::Symbolic)
+function Base.isequal(a::Sym, b::Sym)
+    symtype(a) !== symtype(b) && return false
+    isequal(nameof(a), nameof(b))
+end
+
+function Base.isequal(t1::Term, t2::Term)
     t1 === t2 && return true
-    (istree(t1) && istree(t2)) || return false
+    symtype(t1) !== symtype(t2) && return false
+
     a1 = arguments(t1)
     a2 = arguments(t2)
 
@@ -115,6 +122,7 @@ function Base.isequal(t1::Symbolic, t2::Symbolic)
         length(a1) == length(a2) &&
         all(isequal(l,r) for (l, r) in zip(a1,a2))
 end
+
 ### End of interface
 
 function to_symbolic(x)
@@ -386,7 +394,8 @@ end
 
 Create a term that is similar in type to `t`. Extending this function allows packages
 using their own expression types with SymbolicUtils to define how new terms should
-be created.
+be created. Note that `similarterm` may return an object that has a
+different type than `t`, because `f` also influences the result.
 
 ## Arguments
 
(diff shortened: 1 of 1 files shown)


src/types.jl
```diff
@@ -735,6 +735,29 @@ end
 
 Base.isequal(a::Mul, b::Mul) = isequal(a.coeff, b.coeff) && isequal(a.dict, b.dict)
 
+Base.isequal(b::Add, a::Mul) = isequal(a, b)
+function Base.isequal(a::Mul, b::Add) # this shows up a lot when merging dicts
```
YingboMa, line 739: Maybe this should just be `false`? I think we cannot construct `1 * x^1 * y^1`. @shashi

src/types.jl
```diff
@@ -821,6 +844,14 @@ Base.hash(p::Pow, u::UInt) = hash(p.exp, hash(p.base, u))
 
 Base.isequal(p::Pow, b::Pow) = isequal(p.base, b.base) && isequal(p.exp, b.exp)
 
+Base.isequal(b::Mul, p::Pow) = isequal(p, b)
+function Base.isequal(p::Pow, b::Mul)
```
YingboMa, line 848: So does this, because `x^2 * 1` is just Pow `x^2`.

src/types.jl
shashi: yes,

src/types.jl
shashi: we should never construct `x^2 * 1`, always go directly to `x^2`.

YingboMa:
Test failures are due to some functions break the invariance of the canonical form.

baggepinnen:
Cool!
A further benchmark could be `@time simplify(M[1,1])`

PR #255, opened 2021-03-29
docfix #253

### LINKED ISSUE #253: A typo?
I'm studying the code of this library, starting from `types.jl`. If this is not a good starting point, please let me know.

In the docstring here, https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/142ea18ed2c61ba6f0c1c67d1cc4ddac4fbee54f/src/types.jl#L249-L251
Is it 2 arg `f` or 2 arg `h`?

MasonProtter:
Fixes #253, thank you @linwaytin for pointing this out.


shashi:
Nice.

PR #262, opened 2021-03-31
Simplify `conj` and `real` for `::Real`

### LINKED ISSUE #205: conj(@syms x::Real) ≠ x
I was trying some things out with ModelingToolkit and I ran across this conjugate of Reals, when using the dot product with `Vector{Sym{Real}}`, that is not being simplified

```julia
using SymbolicUtils
@syms x::Real
simplify(conj(x))
```

It would be nice if this simplifies. Not sure how to help with that, the code generation in `methods.jl` is not easy for me.

Suavesito-Olimpiada:
Fix #205

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -4,12 +4,12 @@ import SpecialFunctions: gamma, loggamma, erf, erfc, erfcinv, erfi, erfcx,
                          besselj1, bessely0, bessely1, besselj, bessely, besseli,
                          besselk, hankelh1, hankelh2, polygamma, beta, logbeta
 
-const monadic = [deg2rad, rad2deg, transpose, conj, asind, log1p, acsch,
+const monadic = [deg2rad, rad2deg, transpose, asind, log1p, acsch,
                  acos, asec, acosh, acsc, cscd, log, tand, log10, csch, asinh,
                  abs2, cosh, sin, cos, atan, cospi, cbrt, acosd, acoth, acotd,
                  asecd, exp, acot, sqrt, sind, sinpi, asech, log2, tan, exp10,
                  sech, coth, asin, cotd, cosd, sinh, abs, csc, tanh, secd,
-                 atand, sec, acscd, cot, exp2, expm1, atanh, real, imag, gamma,
+                 atand, sec, acscd, cot, exp2, expm1, atanh, gamma,
                  loggamma, erf, erfc, erfcinv, erfi, erfcx, dawson, digamma,
                  trigamma, invdigamma, polygamma, airyai, airyaiprime, airybi,
                  airybiprime, besselj0, besselj1, bessely0, bessely1]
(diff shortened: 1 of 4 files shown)


Suavesito-Olimpiada:
Fix an additional case where
```julia
@syms x::Real
simplify(real(x)) != x
```
adding `_isreal` to the `utils.jl`.

shashi:
Nice! Since we have also been simplifying "on construction", could you add a method to conj for Symbolic{<:Real} and Symbolic{<:Number}?

shashi (approved):

Suavesito-Olimpiada:
> Nice! Since we have also been simplifying "on construction", could you add a method to conj for Symbolic{<:Real} and Symbolic{<:Number}?

Where should I add these methods? I thought about it but didn't find the correct place. :b

Suavesito-Olimpiada:
Maybe I see, is `methods.jl` the correct file to place this?

Suavesito-Olimpiada:
I think it should be fine by now. :)

src/methods.jl
```diff
@@ -111,8 +111,15 @@ for f in [identity, one, zero, *, +, -]
     @eval promote_symtype(::$(typeof(f)), T::Type{<:Number}) = T
 end
 
+# All of these need to be `Sym` to overwrite the function definitions of
+# the L78, that are made with the type `Sym <: Symbolic`
+#   @number_methods(Sym, term(f, a), term(f, a, b), skipbasics)
 promote_symtype(::typeof(Base.real), T::Type{<:Number}) = Real
-Base.real(s::Symbolic{<:Number}) = islike(s, Real) ? s : term(real, s)
+Base.real(s::Sym{<:Real}) = s
```
shashi, line 118: why is this only `Sym` and not `Symbolic`?

shashi: Oh I just read your comment. Well this seems like it should not be defined in L78, but should be defined outside of the loop there (but could be within `@number_methods` macro). So we should probably remove `real`, `conj` and `imag` from the list of one-argument functions.

test/rulesets.jl
```diff
@@ -18,6 +18,9 @@ end
 
 @testset "Numeric" begin
     @syms a::Integer b c d x::Real y::Number
+    @eqtest simplify(conj(x)) == x
+    @eqtest simplify(real(x)) == x
+    @eqtest simplify(imag(x)) == 0
```
shashi, line 23: You will need to update the tests because `conj(x)` may just return `x`. Use `Term{Real}(conj, [x])` instead.

Suavesito-Olimpiada: Ok, ok.

src/methods.jl
Suavesito-Olimpiada: I'll move them and update the PR.

Suavesito-Olimpiada: Should I force push or add a commit?

shashi: New commit is fine. Either is fine.

PR #265, opened 2021-04-05
Add missing @inbounds in DestructuredArgs

matthieubulte:
Fix for https://github.com/JuliaSymbolics/Symbolics.jl/issues/177

I still need to prepare code changes and a PR to use the `inbounds` flag in
- `Symbolics.jl`
- `ModelingToolkit.jl`

Please let me know if I'm forgetting any other project that have to be updated from this change.

Note that the unit test is a little bit awkward. I tried to find a better way of doing it, but the macro expansion adds the source of the macro, making it hard to use the `test_repr` pattern that is in place. Please let me know if there is a better alternative to what I did.

### DIFF
--- a/src/code.jl
+++ b/src/code.jl
@@ -136,10 +136,11 @@ end
     elems
     inds
     name
+    inbounds::Bool
 end
 
-function DestructuredArgs(elems, name=gensym("arg"); inds=eachindex(elems))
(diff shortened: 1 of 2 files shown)


YingboMa (approved):
LGTM

shashi:
Looks good! Thanks.

PR #266, opened 2021-04-06
Extend and document canonical form of `Symbolic{<:Number}`

### LINKED ISSUE #256: Document canonical form of Number terms
We have an informal "canonical" form for Terms of symtype Number right now. But we don't have a good succinct definition of the same.

For example `1 * x` always returns `x`. `x * x` always returns `x^2` and so on.

Having a definition of this will help us speed up things like `isequal` and have it be more accurate.

Suavesito-Olimpiada:
fix #256

### DIFF
--- a/page/index.md
+++ b/page/index.md
@@ -151,7 +151,20 @@ SymbolicUtils contains [a rule-based rewriting language](/rewrite/#rule-based_re
 
 By default `*` and `+` operations apply the most basic simplification upon construction of the expression.
 
-Commutativity and associativity are assumed over `+` and `*` operations on `Symbolic{<:Number}`.
+The rules with which the canonical form of `Symbolic{<:Number}` terms are constructed are the next (where `x isa Symbolic` and `c isa Number`)
+
+ -  `0 + x` and `1 * x` always gives `x`
+ -  `0 * x` always gives `0`,
+ -  commutativity and associativity over `+` and `*` are assumed. Re-ordering of terms will be done under a [total order](https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/master/src/ordering.jl)
+ -  sum of `Add`'s are fused
+ -  product of `Mul`'s are fused
+ -  `x + ... + x` will be fused into `n*x` with type `Mul`
+ -  `x * ... * x` will be fused into `x^n` with type `Pow`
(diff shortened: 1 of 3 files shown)


Suavesito-Olimpiada:
As I saw when writing this docs, there are things that can be done to further improve the canonical form just using `Mul`, `Add` and `Pow`. Some of these are "fussing" expressions of type `Pow` and `Mul`.

I can contribute these this week, but I don't know if put them in this PR or make another with "just the code". If is needed in another PR i would extend the docs here to reflect the changes made in the other PR.

I also don't know if I should add here in the docs the optimizations for some functions (as in [262](https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/262)) or not, and all the "basic" operators `[+, -, \, /, ^, *]`.

shashi:
> Some of these are "fussing" expressions of type Pow and Mul.

What do you mean by this?

What are the changes you're planning to make? Feel free to put them in this PR! :)

Suavesito-Olimpiada:
> > Some of these are "fusing" expressions of type Pow and Mul.
> 
> What do you mean by this?

In master now you have that

```julia
julia> x*inv(x)
1
```

For some `x` expressions, but

```julia
julia> (x^-1)*inv(x^-1)
(x^-1)*((x^-1)^-1)
```

And you can see why this is happening

```julia
julia> print_tree((x^-1)*inv(x^-1))
*
├─ 1
├─ (^)
│  ├─ x
│  └─ -1
└─ (^)
   ├─ ^
   │  ├─ x
   │  └─ -1
   └─ -1
```

Basically `Mul` is not merging (what I called "fusing") with `Pow`.

Suavesito-Olimpiada:
I found what it is happening, the `Mul` *does merge with `Pow`, but `Pow` doesn't reduce with another `Pow`.

Suavesito-Olimpiada:
Now with this PR, aside the documentation

```julia
julia> (x^-1)*inv(x^-1)
1
```

Still needs to finish the documentation part.

YingboMa:
Could you add a test for `(x^-1)*inv(x^-1)` in here https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/491ad806c8a21b75c11efd930fbce7af0de1fcfe/test/basics.jl#L187

Suavesito-Olimpiada:
Yeah, sure. I finished adding the documentation. And I added the test.

Suavesito-Olimpiada:
Locally all the test passed :smile:, now to the next issue. :stuck_out_tongue:

PR #267, opened 2021-04-06
Fix polynorm deprecation warning

david-pl:
Currently errors with `ERROR: MethodError: no method matching depwarn(::String)`


shashi:
Thank you! :raised_hands:

PR #271, opened 2021-04-08
Allow any implementation of the MultivariatePolynomials API

blegat:
The current implementation only supports the `MPoly` polynomial representation of AbstractAlgebra.
By using the MultivariatePolynomials, any implementation of multivariate polynomials can be used as long as it implements the API.
I choose to add a default with DynamicPolynomials which has a reasonable tradeoff between compile-time vs runtime.
On the other hand, [TypedPolynomials](http://github.com/JuliaAlgebra/TypedPolynomials.jl/) is usually faster but its compile time can become prohibitive if there are too many variables so it's best no to use it by default.
I added a benchmark.
Before this PR:
```
julia> goldstein_price();
  2.088 ms (19781 allocations: 855.57 KiB)
  6.695 ms (80770 allocations: 3.36 MiB)
  15.477 ms (202060 allocations: 8.39 MiB)
```
After this PR:
```
julia> goldstein_price();
  1.104 ms (7483 allocations: 394.23 KiB)
  2.431 ms (16981 allocations: 1019.19 KiB)
  6.018 ms (36690 allocations: 2.44 MiB)
```
The representation used by DynamicPolynomials is mostly similar to `AbstractAlgebra.MPoly` so I guess the different comes from the fact that `AbstractAlgebra` used BigInt while DynamicPolynomials automatically promotes depending on what you multiply it with (so `Int` in this case).
Another difference is that even if you create, a lot of variables, the monomials of DynamicPolynomials won't contain all of them.
There is not a ring ZZ like in AbstractAlgebra which forces all monomials to have many exponents. In DynamicPolynomials and TypedPolynomials, you automatically get more variables as you multiply and add things together but it starts very sparse.
I don't think we notice a difference in this benchmark but it might get noticeable as
https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/491ad806c8a21b75c11efd930fbce7af0de1fcfe/src/abstractalgebra.jl#L85-L86
will typically create a lot of variables with subexpressions only involving a small subset of them.

Edit: fixes #271


shashi:
A few things I wanted to know more about:

- Going forward we're going to need GCD see #231 
- Does DynamicPolynomials.jl or one of the packages this would enable use dense array polynomial representation?
- Does this leave the whole system non-GPL? (MIT / BSD licenses are ok)
- Would it be a lot of work to also allow swapping in AbstractAlgebra.MPoly in place of these polynomials?

blegat:
> Going forward we're going to need GCD see #231

I noticed you opened an issue some time ago about this (https://github.com/JuliaAlgebra/DynamicPolynomials.jl/issues/57) but we never replied, sorry about that. `gcd` is only implemented for univariate polynomials at the moment but we can add it for multivariate polynomials as well. By the way, this is also a blocking issue for simplifying our rational function (see https://github.com/JuliaAlgebra/MultivariatePolynomials.jl/blob/a58fa070cf4d82943702f0366b3c192cc2995cb5/src/rational.jl#L3).

> Does DynamicPolynomials.jl or one of the packages this would enable use dense array polynomial representation?

What do you mean by dense exactly ? Do you mean we store the coefficients for all the monomials up to some degree, even if they are zero (hence we don't need to store the exponents ?). This is implemented in https://github.com/JuliaDiff/TaylorSeries.jl with the `TaylorN` types which contains on `HomogeneousPolynomial` for each degree which contains just a vector of coefficients (and not the corresponding monomials since it's all of them). `TaylorN` does not implement the MultivariatePolynomials API but it wouldn't be hard to do if we focus on the part of the API needed for SymbolicUtils.

> Does this leave the whole system non-GPL? (MIT / BSD licenses are ok)

Yes, there are very few dependencies (MutableArithmetics, OrderedCollections and DataStructures) which are all licensed under the MIT license.

> Would it be a lot of work to also allow swapping in AbstractAlgebra.MPoly in place of these polynomials?

No, we just need to implement the MultivariatePolynomials API for `AbstractAlgebra.MPoly`. MultivariatePolynomials is really just an API for allowing software to use multivariate polynomials without having to rely on a particular representation/implementation. Using this API allows SymbolicUtils to support essentially any implementation at the cost of just a lightweight depencency. They you can switch the default at any time and the user can play around different implementation, finding the one that best suits its use case.

PR #273, opened 2021-04-09
WIP : Partially implemented a linear coefficient checker

SamuelzXu:
A partially completed a function that tells whether an expression is linear in any of a list of variables. Any feedback is appreciated.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -220,7 +220,7 @@ const monadic_linear = [deg2rad, +, rad2deg, transpose, -, conj]
 const monadic_nonlinear = [asind, log1p, acsch, erfc, digamma, acos, asec, acosh, airybiprime, acsc, cscd, log, tand, log10, csch, asinh, airyai, abs2, gamma, lgamma, erfcx, bessely0, cosh, sin, cos, atan, cospi, cbrt, acosd, bessely1, acoth, erfcinv, erf, dawson, inv, acotd, airyaiprime, erfinv, trigamma, asecd, besselj1, exp, acot, sqrt, sind, sinpi, asech, log2, tan, invdigamma, airybi, exp10, sech, erfi, coth, asin, cotd, cosd, sinh, abs, besselj0, csc, tanh, secd, atand, sec, acscd, cot, exp2, expm1, atanh]
 
 # Check for linear coefficients
-coefficients(x::Term, expr) = coefficients([Term=>true],expr)
+coefficients(x::Term, expr) = coefficients([x=>true],expr)
 function coefficients(f::Dict,expr)
     terms = keys(f)
     res = Dict{Any, Bool}
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -227,11 +227,11 @@ function coefficients(expr,syms::AbstractArray{SymbolicUtils.Sym})
 end
 
 function has_nonlinear(expr,sym,above)
-    if istree(expr)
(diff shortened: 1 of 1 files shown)


SamuelzXu:
@YingboMa

src/utils.jl
```diff
@@ -214,3 +214,40 @@ macro matchable(expr)
         Base.length(x::$name) = $(length(fields) + 1)
     end |> esc
 end
+
+# Check for linear coefficients
+coefficients(expr,sym::SymbolicUtils.Sym) coefficients(expr,[sym])
```
YingboMa, line 219: The signature should be `coefficients(f, expr)`

src/utils.jl
```diff
@@ -214,3 +214,40 @@ macro matchable(expr)
...
+function coefficients(expr,syms::AbstractArray{SymbolicUtils.Sym})
```
YingboMa, line 220: Please follow the coding style of the rest of the package. Add a space after a comma.

src/utils.jl
```diff
@@ -214,3 +214,40 @@ macro matchable(expr)
...
+    res = Dict{SymbolicUtils.Sym,Bool}
```
YingboMa, line 221: It should be `Dict{Any,Bool}`, we shouldn't just restrict this to Sym. We can have `coefficient(a*sin(x), sin(x)) => a`

src/utils.jl
```diff
@@ -214,3 +214,40 @@ macro matchable(expr)
...
+    for s in syms
+        cur = Dict(s,has_nonlinear(expr,s,false))
+        merge(res,Dict)
```
YingboMa, line 224: This should just be `res[s] = has_nonlinear(expr,s,false)` instead of creating a new dict in every iteration

src/utils.jl
```diff
@@ -214,3 +214,40 @@ macro matchable(expr)
...
+    end
+    return res
+end
+
+function has_nonlinear(expr,sym,above)
+    if SymbolicUtils.istree(expr)
+        if !above
+            is_nonlinear(expr) && above=true
```
YingboMa, line 232: This is not valid syntax. There should be parentheses `xxx && (a = 1)`

src/utils.jl
```diff
@@ -214,3 +214,40 @@ macro matchable(expr)
...
+end
+
+function is_nonlinear(expr)
+    op = SymbolicUtils.operation(expr)
+    if isa(op,Add)
+        # need clarifications on this
+    elseif isa(op,Mul)
+    elseif isa(op,Term)
+    return
+end
```
shashi, line 253: There's a pretty accurate linearity checker in https://github.com/JuliaSymbolics/Symbolics.jl/blob/master/src/linearity.jl

shashi (changes requested):
I would rather have this be in Symbolics.jl since it's not about representation and rewriting.

src/utils.jl
shashi: Seems like wrong syntax too.

src/utils.jl
SamuelzXu: Awesome! Just what I was looking for!

SamuelzXu:
@YingboMa Any suggestions? Also, about what @shashi mentioned with transferring this to Symbolics, what do you think?

YingboMa:
Closed by https://github.com/JuliaSymbolics/Symbolics.jl/pull/248

PR #277, opened 2021-04-12
Add `@numrule` for rules matching `[-, /, \, \\] expressions`

### LINKED ISSUE #241: Pattern Matching with subtraction is seemingly broken
```julia
julia> @acrule(-1 - ~x => "match")(-1 - x)

julia> @rule(-1 - ~x => "match")(-1 - x)

```
ref https://julialang.zulipchat.com/#narrow/stream/274208-helpdesk-(published)/topic/Symbolics.2Ejl.20derivative.20cot(x)

Suavesito-Olimpiada:
fix #241 adding `@numrule` as proposed by @shashi

### DIFF
--- a/src/SymbolicUtils.jl
+++ b/src/SymbolicUtils.jl
@@ -27,7 +27,7 @@ include("rewriters.jl")
 using .Rewriters
 
 using Combinatorics: permutations, combinations
-export @rule, @acrule, RuleSet
+export @rule, @acrule, @numrule, RuleSet
 
 # Rule type and @rule macro
 include("rule.jl")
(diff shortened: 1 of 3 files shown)


shashi:
This is pretty awesome!! Thank you!

shashi:
Can we try doing these in `simplify_rules.jl`?

Suavesito-Olimpiada:
Yeah! of course, I just wanted to have some insight on the aproach and the implementation, so f it was OK then I could worked it further.

If you think this is fine, I'll get my hand into `symplify_rules.jl`. :smile:

Suavesito-Olimpiada:
Coming back to check this. Now it seems even messier than last time, any commentaries?

shashi:
Yeah since then Div has been introduced. It should actually help in this case. We still badly need this feature!

PR #278, opened 2021-04-13
Add adjoint for `create_array`

### LINKED ISSUE #274: Define Zygote adjoint of `create_array`
I think this is the more robust solution to https://github.com/SciML/DataDrivenDiffEq.jl/issues/54.

I'm open to adding this patch behind a `@requires`.

cc @DhairyaLGandhi

DhairyaLGandhi:
Added an adjoint as per discussion in #274 

Closes #274


test/adjoints.jl
```diff
@@ -0,0 +1,28 @@
+using SymbolicUtils, SymbolicUtils.Code
+using Zygote
```
shashi, line 2: Can you add this to the test dependencies?

test/adjoints.jl
```diff
@@ -0,0 +1,28 @@
...
+    y = SymbolicUtils.Code.create_array(A, T, Val(u), Val(dim), elems...)
+    @test y == ŷ
+
+    gs = pb(ones(T, length(elems)))
+    @test length(gs[1]) == length(elems)
+    for i = 1:(prod(dim)-1)
+      @test gs[1][i] == one(eltype(ŷ))
+    end
+  end
+end
```
shashi, line 28: @AlCap23 and @doddgray would love to know if this is correct and works for your case.

test/adjoints.jl
shashi: Ah I see it, sorry missed that. :+1:

shashi (approved):

YingboMa (approved):

test/adjoints.jl
doddgray: @shashi sorry for my delayed response. Thank you for checking in! So far Zygote gradients involving array-valued functions generated with `Symbolics.build_function(A::AbstractArray{Num},args...)` now seem to work and provide correct results, for my test cases. I still need to figure out how to properly use SymbolicUtils' code generation before I can take advantage. I see reasonable primal and Zygote.gradient evaluation performance when I generate a functions in the global scope, but functions which run `build_function` internally (using `expression=Val{false}` for `RuntimeGeneratedFunctions`) have very inefficient primals and gradients. I don't think this is a Zygote-specific issue because the primal evaluations suffer similarly to the gradients, rather I think the generated functions are being compiled over and over again?  Can you recommend a way to initialize/compile RGF's that will be evaluated many times used within a function? I ask here because I worry that Zygote performance might depend on the scope or module where the function is generated, but maybe not the right place...

PR #279, opened 2021-04-13
Document the behavior Walk with bare rewriters

adamslc:
Walk really needs the rewriter it works with to have the PassThrough
property. Chain also has this property.

xref #272.


shashi:
Looks great!!

PR #284, opened 2021-04-19
Fix substitute when using metadata

### LINKED ISSUE #283: Metadata of Term changes in substitution
```julia
using SymbolicUtils
using SymbolicUtils: Term
@syms a b f t
f = Term(f, (t,)) # create term f(t)
f = setmetadata(f, Bool, true)
hasmetadata(f, Bool) # true
newf = substitute(f, Dict(a=>b)) # unrelated substitution
hasmetadata(newf, Bool) # false
```
I think this is a M(not)WE for `SymbolicUtils`. The problem came up after recent changes to MTK SciML/ModelingToolkit.jl/pull/943
```julia
using ModelingToolkit
using ModelingToolkit: isparameter
@parameters t a(t)
@variables x(t) y(t)
isparameter(a) # true
isparameter(substitute(a, x=>y)) # false :(
```

shashi:
fixes #283

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -62,7 +62,7 @@ Get all the metadata of a term or `nothing` if no metadata is defined.
 
 """
 metadata(s::Symbolic) = s.metadata
-metadata(s::Any) = x
+metadata(s::Any) = nothing
 
 function hasmetadata(s::Symbolic, ctx)
     metadata(s) isa AbstractDict && haskey(metadata(s), ctx)
--- a/src/types.jl
+++ b/src/types.jl
@@ -56,8 +56,16 @@ symtype(x) = typeof(x)
 
 symtype(::Symbolic{T}) where {T} = T
 
+"""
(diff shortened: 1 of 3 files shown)


YingboMa (approved):

src/types.jl
```diff
@@ -56,8 +56,16 @@ symtype(x) = typeof(x)
 
 symtype(::Symbolic{T}) where {T} = T
 
+"""
+    metadata(s)
+Get all the metadata of a term or `nothing` if no metadata is defined.
+
+"""
+metadata(s::Symbolic) = s.metadata
+metadata(s::Any) = x
```
adamslc, line 65:
```suggestion
metadata(s::Any) = nothing
```

adamslc (changes requested):
One typo to fix.

shashi:
@adamslc thanks!

PR #286, opened 2021-04-23
Allow for MArrays

RafaelArutjunjan:
Currently, `create_array()` is only implemented for `SArray`s but errors for `MArray`s.

### DIFF
--- a/src/code.jl
+++ b/src/code.jl
@@ -464,6 +464,15 @@ end
     SArray{Tuple{dims...}, T}(elems...)
 end
 
+## MArray
+@inline function create_array(::Type{<:MArray}, ::Nothing, nd::Val, ::Val{dims}, elems...) where dims
+    MArray{Tuple{dims...}}(elems...)
+end
+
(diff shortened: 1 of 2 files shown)


YingboMa (commented):
Could you add tests?

RafaelArutjunjan:
Given that the `===` operator does not work with `MArray`s in the same way as for `SArray`s, i.e. `(@SVector [1, 2, 3, 1/2]) === (@SVector [1, 2, 3, 1/2])` evaluates as `true` but `(@MVector [1, 2, 3, 1/2]) === (@MVector [1, 2, 3, 1/2])` is `false` (I guess due to the whole non-allocation thing with `SArray`) I went for
```julia
R1 = eval(toexpr(Let([a ← 1, b ← 2, arr ← @MVector([1,2])],MakeArray([a,b,a+b,a/b], arr))))
@test R1 == (@MVector [1, 2, 3, 1/2]) && R1 isa MVector
```
A little clumsy but it works. Feel free to change it if you have a better idea.

shashi:
Yeah that test looks good! Thanks for the nice PR!

PR #287, opened 2021-04-23
More robust subtype check

YingboMa:
Sometimes Julia turn `Tuple{Any}` into `Tuple{T} where T`, so we cannot reliably use `T.parameters` to query type parameters.
```julia
julia> SymbolicUtils.Sym{T}
Sym{SymbolicUtils.FnType{Tuple{Any}, Real}, M} where M

julia> SymbolicUtils.Sym{T}(:t) |> dump
Sym{SymbolicUtils.FnType{Tuple{T} where T, Real}, Nothing}
  name: Symbol t
  metadata: Nothing nothing
```

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -215,18 +215,11 @@ function promote_symtype(f::Sym{FnType{X,Y}}, args...) where {X, Y}
         return Y
     end
 
-    nrequired = fieldcount(X)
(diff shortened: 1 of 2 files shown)


shashi:
okay I guess...

shashi (approved):

PR #298, opened 2021-06-09
don't rely on isless to be defined on elements, use hash as fallback

shashi:
useful when dealing with `ArrayOp`s


YingboMa (approved):

shashi:
Tests pass for me  locally dunno what's happening here.

PR #299, opened 2021-06-12
Don't use testset in fuzz.jl

YingboMa:
It turns out that at testset macro sets the random seed


shashi:
Can't we just set the seed inside the testset?

shashi:
bump

PR #302, opened 2021-06-18
TermInterface.jl integration

0x0f0f0f:
Integrating (defining methods) for https://github.com/JuliaSymbolics/TermInterface.jl which should serve as a glue package providing an interface for common functions to manipulate symbolic expressions (not exclusively SymbolicUtils.jl) between various packages like Metatheory.jl

Opening PR to keep track. Not ready to be merged


page/interface.md
```diff
@@ -18,25 +20,25 @@ SymbolicUtils matchers can match any Julia object that implements an interface t
...
 
 Check if `x` represents an expression tree. If returns true,
-it will be assumed that `operation(::T)` and `arguments(::T)`
+it will be assumed that `gethead(::T)` and `getargs(::T)`
 methods are defined. Definining these three should allow use
 of `simplify` on custom types. Optionally `symtype(x)` can be
 defined to return the expected type of the symbolic expression.
 
-#### `operation(x::T)`
+#### `head(x::T)`
```
YingboMa, line 31: head or gethead?

YingboMa (commented):
Could you also merge master into this branch and update the unsorted_arguments as well?

page/interface.md
0x0f0f0f: gethead. fixing

0x0f0f0f:
> Could you also merge master into this branch and update the unsorted_arguments as well?

will do ASAP

0x0f0f0f:
I can't seem to understand why codegen is broken in `code.jl`!! @shashi @YingboMa I need some help here :( 

It broke when i changed definition of `isterm` in TermInterface.jl to:
```
"""
    isterm(x)

Returns `true` if `x` is a term. If true, `gethead`, `getargs`
must also be defined for `x` appropriately.
"""
isterm(x) = isterm(typeof(x))
isterm(x::Type{Expr}) = true
isterm(x::Type{T}) where {T} = false
export isterm
```

shashi:
Would it be possible to not use this definition on Expr? Because I'm not sure we will be able to support all the interface like symtype and metadata on Expr.

0x0f0f0f:
Fixed the codegen bug, just by adding
```julia
toexpr(O::Expr, st) = O
```
to `code.jl` :)

0x0f0f0f:
> Would it be possible to not use this definition on Expr? Because I'm not sure we will be able to support all the interface like symtype and metadata on Expr.

@shashi it makes sense, but I think that there should be default limited behaviour on `Expr`. Other packages, for example MT, should have default behaviour for Expr. Let me know. Tests are now passing

0x0f0f0f:
So, in order to go ahead with integrating MT with SU and Symbolics, here's the ops plan:

1. (Optional but strongly suggested for interoperability). Separate Rewriters into different package. Only depends on TermInterface.jl #341 
2. (Alternatively) Merge this PR first and release new SymbolicUtils version (keeping an eye on compat)
3. Metatheory.jl can now depend on Rewriters/SymbolicUtils.jl and TermInterface.jl. Everything will be pulled from registry. No packages added through WIP branches. MT new version can be released.
4. Step 2 can also be done now
5. Symbolics.jl can now depend on TermInterface.jl, new SymbolicUtils.jl version and new MT version. No more WIP branches. Can be released.
6. Finish integrating MT optimization in Symbolics.jl
7. Update ModelingToolkit.jl, Catalyst.jl, NeuralPDE.jl etc...

Also pinging @ChrisRackauckas

0x0f0f0f:
🚀

PR #318, opened 2021-08-03
WIP: `propagate_metadata`

shashi:
```julia
julia> import Symbolics: VariableSource
julia> SymbolicUtils.propagate_metadata(::Type{VariableSource}, ::typeof(sin), x) = getmetadata(x,VariableSource)
julia> @variables x
1-element Vector{Num}:
 x
julia> sin(x.val).metadata
Base.ImmutableDict{DataType, Any} with 1 entry:
  VariableSource => (:variables, :x)
```

- [X] `Term`
- [ ] `Add`
- [ ] `Mul`
- [ ] `Pow`
- [ ] `(::Sym{<:FnType})(...)`...?
- [ ] benchmark

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -80,6 +80,32 @@ function getmetadata(s::Symbolic, ctx, default)
     s.metadata isa AbstractDict ? get(s.metadata, ctx, default) : default
 end
 
+propagate_metadata(T, f, args...) = nothing
+
+function _propagate_metadata(f, args)
+    if any(x->metadata(x) !== nothing, args)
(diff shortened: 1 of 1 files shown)


src/types.jl
```diff
@@ -388,13 +414,13 @@ function _promote_symtype(f, args)
     end
 end
 
-function term(f, args...; type = nothing)
+function term(f, args...; type = nothing, metadata = _propagate_metadata(f, args))
```
shashi, line 417: IN THEORY! The `any` check should get compiled away, at least here.

PR #329, opened 2021-08-06
Add where syntax to rules

### LINKED ISSUE #322: Predicate over multiple slot variables
In the documentation, an example is given to have a predicate check a slot variable before matching. For a problem that I'm working on, I would like to check multiple slot variables. For example, check whether, for slot variables `~x` and `~y`, their product is smaller than four:

```
@syms x y

mapping = Dict(:x => 2, :y => 3)

below_four(x, y) = mapping[x] * mapping[y] < 4

r = @rule (~x + ~y)::below_four => ~x
```

I'm aware that this is a bit of a weird situation, because the application of the rule depends on state outside of the expression. However, it's a real world use-case and I didn't make it up :yum: No one said that expressions should be pure.

rikhuijzer:
Closes #322.

@shashi, I think that it would be good to also test a more complicated rule with the predicate, but I couldn't get it to work. Could you either add it to the PR or give a suggestion?

### DIFF
--- a/src/rule.jl
+++ b/src/rule.jl
@@ -154,9 +154,9 @@ The `:where` is rewritten from, for example, `~x where f(~x)` to `f(~x) ? ~x : n
 """
 function rewrite_rhs(expr::Expr)
     if expr.head == :where
-        left = expr.args[1]
-        right = expr.args[2]
-        expr = Meta.parse("$right ? $left : nothing")
+        rhs = expr.args[1]
+        predicate = expr.args[2]
+        expr = Meta.parse("$predicate ? $rhs : nothing")
     end
     return expr
 end
--- a/src/rule.jl
+++ b/src/rule.jl
@@ -147,6 +147,22 @@ function (r::Rule)(term)
 end
 
 """
+    rewrite_rhs(expr::Expr)
+
+Rewrite the `expr` by dealing with `:where` if necessary.
+The `:where` is rewritten from, for example, `~x where f(~x)` to `f(~x) ? ~x : nothing`.
+"""
+function rewrite_rhs(expr::Expr)
+    if expr.head == :where
+        left = expr.args[1]
+        right = expr.args[2]
+        expr = Meta.parse("$right ? $left : nothing")
+    end
+    return expr
+end
+rewrite_rhs(expr) = expr
+
+"""
     @rule LHS => RHS
 
 Creates a `Rule` object. A rule object is callable, and  takes an expression and rewrites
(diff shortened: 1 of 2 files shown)


test/rulesets.jl
```diff
@@ -131,3 +131,20 @@ end
...
+
+@testset "where" begin
+    expected = Meta.parse("f(~x) ? ~x + ~y : nothing")
+    @test SymbolicUtils.rewrite_rhs(:((~x + ~y) where f(~x))) == expected
+
+    @syms a b
+    f(x) = x === a
+    r = @rule ~x => ~x where f(~x)
+    @eqtest r(a) == a
+    @test isnothing(r(b))
```
YingboMa, line 143: `isnothing` is slower than `=== nothing`

rikhuijzer: Ok, I'll change this. Please do note that there are currently 5 occurrences of `isnothing` in `src` and 3 in `test/rulesets.jl`. Unlike it's usage in tests, the occurrences in `src/` will affect production performance. Also, due to https://github.com/JuliaLang/julia/pull/38905, the performance difference between `x === nothing` and `isnothing(x)` got even smaller.

rikhuijzer: The linked issue only got into Julia 1.7 as far as I can see, so you probably have a good point given that Julia 1.6 is still very likely to become LTS

src/rule.jl
```diff
@@ -221,7 +237,8 @@ julia> r(2 * (a+b+c))
 
 **Predicates**:
 
-Predicates can be used on both `~x` and `~~x` by using the `~x::f` or `~~x::f`.
+There are two kinds of predicates, namely over slot variables and over the whole rule.
+For the former, predicates can be used on both `~x` and `~~x` by using the `~x::f` or `~~x::f`.
```
shashi, line 241: I'm thinking we should maybe deprecate `~x::f`...

It's kind of a confusing syntax, compare:

```julia
@rule ~x::(x->x isa Integer) => 0
@rule ~x => 0 where ~x isa Integer
```

@YingboMa ?

shashi: Oh never mind actually it is more efficient to use `::` because it stops parsing preemptively.

shashi (approved):

rikhuijzer:
Hey xD I'm still processing the very nice tip to use `:($x ? $y : nothing)` instead of `Meta.parse`

shashi:
Oh no, I thought it's only in the tests. Sorry, would you mind opening another PR?

rikhuijzer:
> Oh no, I thought it's only in the tests. Sorry, would you mind opening another PR?

Np. Will be there in a few mins.

src/rule.jl
```diff
@@ -146,6 +146,22 @@ function (r::Rule)(term)
...
+    rewrite_rhs(expr::Expr)
+
+Rewrite the `expr` by dealing with `:where` if necessary.
+The `:where` is rewritten from, for example, `~x where f(~x)` to `f(~x) ? ~x : nothing`.
+"""
+function rewrite_rhs(expr::Expr)
+    if expr.head == :where
+        rhs = expr.args[1]
+        predicate = expr.args[2]
+        expr = Meta.parse("$predicate ? $rhs : nothing")
```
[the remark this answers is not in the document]
rikhuijzer: Fixed in #335.

PR #333, opened 2021-08-09
Add a doctest

rikhuijzer:
This is just a suggestion. In the Julia library, Documenter.jl is used to verify all the documentation via doctests. This means that Documenter's doctest logic should be quite robust. Since this package has a lot of examples in the docstrings, maybe it is a good idea to run doctests to avoid broken documentation.

EDIT: For example, other packages which also use doctests are [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl) and [MCMCChains.jl](https://github.com/TuringLang/MCMCChains.jl).

### DIFF
--- a/.github/workflows/DeployPage.yml
+++ b/.github/workflows/DeployPage.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Install Julia
       uses: julia-actions/setup-julia@v1
       with:
-        version: 1.5
+        version: 1 # Latest stable Julia version.
     # NOTE
     # Adjust the `input=""` on the last line to indicate where
     # the source of the package page is (`page/` by default).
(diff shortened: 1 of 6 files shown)


shashi:
Nice!

How does jldoctest get rendered with Franklin? btw The SymbolicUtils documentation is generated with Franklin... We may have to hack https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/master/page/utils.jl#L5-L26 if it doesn't already render the code block properly.

rikhuijzer:
> How does jldoctest get rendered with Franklin?

I couldn't run the SymbolicUtils build locally due to some broken dependency (I'm running NixOS where I get those issues from time to time). Therefore, I've tested it in Franklin.jl and there the output is indeed incorrect for `jldoctest` code blocks.

The workaround is added in `d3423d2`, but note that I **didn't test it**.

shashi:
> The workaround is added in d3423d2

Looks good enough.

shashi:
Thanks! This is great!

PR #336, opened 2021-08-10
fix vect with Polynomials

shashi:
ref https://github.com/JuliaAlgebra/MultivariatePolynomials.jl/issues/169


YingboMa (approved):
Pretty good! This doesn't do type piracy at lease.

PR #337, opened 2021-08-11
Add sort kwarg in arguments to avoid extra computation

YingboMa:
@shashi what do you think about this? It might be a little breaking in the API sense.


shashi:
This is part of the term interface. Kwarg might make that confusing. Can we just use a different function? `unsorted_arguments` could just call `arguments` by default but special case for Add and Mul...

YingboMa:
We should use `unsorted_arguments` more.

Issue #342, opened 2021-08-24
Consider adding a @capture macro that injects matches into the current scope.


peterahrens:
Such a macro could be useful for rules with more complicated consequents. Considering the following (untested) code snippet which defines a rule to sort the terms of a polynomial by their power. Perhaps there's a cleaner way to do this, but I personally find it easier to use `@capture` for nested rule processing or for non-rule tasks, like compilation or analysis.
```
function my_complicated_rule(f)
if @capture f *(~~ts)
  poly_terms = []
  for t in ts
    (@capture t ~x ^ ~p) || return nothing
    push!(poly_terms, (p, x))
  end
end
sort!(poly_terms, reverse=true)
return mapreduce(((p, x),)->x^p, *, poly_terms)
```

PR #343, opened 2021-08-24
add capture macro

peterahrens:
This commit adds a simple macro for simply matching terms and injecting the matches into the current scope, as described in #342. The macro returns true or false depending on whether the pattern was matched.


peterahrens:
Alright, docstring and tests added!

src/rule.jl
```diff
@@ -316,6 +316,38 @@ macro rule(expr)
     end
 end
 
+"""
+    @capture ex LHS
```
MasonProtter, line 320:
```suggestion
    @capture ex pattern
```
I think instead of calling it LHS here, it'd be better to call it `pattern`, since there isn't really a RHS!

src/rule.jl
```diff
@@ -316,6 +316,38 @@ macro rule(expr)
...
+
+Uses a `Rule` object to capture an expression if it matches the LHS pattern. Returns `true` and injects
+slot variable match results into the calling scope when the LHS matches, otherwise returns false. The
+rule language for specifying the LHS pattern is the same in @capture as it is in `@rule`. Contextual matching
+is not yet supported
```
MasonProtter, line 325:
```suggestion
Uses a `Rule` object to capture an expression if it matches the `pattern`. Returns `true` and injects
slot variable match results into the calling scope when the `pattern` matches, otherwise returns false. The
rule language for specifying the `pattern` is the same in @capture as it is in `@rule`. Contextual matching
is not yet supported

e.g.

    julia> @syms a; ex = a ^ a;

    julia> if @capture ex (~x) + (~x)
               @show x
           elseif @capture ex (~y) ^ (~y)
               @show y
           end;
    y = a
```
I think adding an example like this might help motivate the idea behind the macro, and show people why it's useful for it to return a `Bool`. Not sure if this is an ideal example, but it's the first that came to mind.

src/SymbolicUtils.jl
```diff
@@ -35,7 +35,7 @@ include("rewriters.jl")
 using .Rewriters
 
 using Combinatorics: permutations, combinations
-export @rule, @acrule, RuleSet
+export @rule, @acrule, RuleSet, @capture
```
MasonProtter, line 38: One thing to think about is if there are any packages that do
```julia
using MacroTools, SymbolicUtils
```
If so, this addition will be a breaking change for them. Not sure that's really worth the trouble of not adding this functionality, but I thought I'd bring it up.

MasonProtter (commented):
Looks great @peterahrens! Here are a couple random thoughts.

src/SymbolicUtils.jl
peterahrens: We could decide not to export, or call it `@match` instead perhaps?

MasonProtter: `@match` also exists elsewhere so it doesn’t really help. I think it’s probably fine to export, but we should bump the version.

peterahrens:
Alright! Thanks @MasonProtter for the suggestions! I've resolved everything except for the potential macrotools name collision. It might be easiest to simply leave the macro un-exported, but I'll wait for another opinion on the matter.

peterahrens:
The version has been bumped to 16.0 because this is a breaking change.

shashi:
I don't see any breaking changes though? What's something that doesn't work that used to work?

peterahrens:
`using MacroTools; using SymbolicUtils` used to work, now results in name collision because we export `@capture`. If you don't want me to bump the version I can set it back.

shashi:
Ah ok!

YingboMa:
We really shouldn't export `@capture`
```julia
[ Info: Precompiling ModelingToolkit [961ee093-0014-501f-94e3-6117800e7a78]
WARNING: both MacroTools and SymbolicUtils export "@capture"; uses of it in module Symbolics must be qualified
WARNING: both Symbolics and MacroTools export "@capture"; uses of it in module ModelingToolkit must be qualified
```

peterahrens:
Perhaps that should go in a separate PR? Unless you'd like me to push a quick fix to this branch, which I'm willing to do.

YingboMa:
Yeah, let's make a new PR and a new breaking release.

PR #346, opened 2021-08-26
Optimized ordering computation

YingboMa:


test/order.jl
```diff
@@ -81,8 +81,7 @@ end
 @testset "small terms" begin
     # this failing was a cause of a nasty stackoverflow #82
     @syms a
-    @test Term(^, [a, -1]) <ₑ (a + 2)
-    @test !((a + 2) <ₑ Term(^, [a, -1]))
+    istotal(Term(^, [a, -1]), (a + 2))
```
YingboMa, line 84: A consequence of this that `Term(^, [a, -1])` is different from `Pow(a, -1)`. @shashi what do you think?

shashi: You  mean they aren't `isequal`? I think that's fine.

shashi:
let's backport this to 0.13.x

PR #353, opened 2021-09-02
speed up `flatten_fraction`

shashi:


shashi:
Lol it looks like the old thing just took way too long? So strange!!

PR #360, opened 2021-09-10
Updated TermInterface.jl

0x0f0f0f:
Added `exprhead` that is used by MT's pattern matcher to match against the `<:Pattern` types. Fundamental in order to make the e-graph rewriting of symbolics to work.
Added `unsorted_arguments` to TermInterface

Now default interface for `Expr` is polished. TODO: add documentation saying that using `Expr` as Symbolic types is not probably a good idea.

### DIFF
--- a/test/interface.jl
+++ b/test/interface.jl
@@ -1,12 +1,6 @@
 using SymbolicUtils, Test
 using TermInterface
 
-TermInterface.istree(ex::Expr) = ex.head == :call
-TermInterface.operation(ex::Expr) = ex.args[1]
-TermInterface.arguments(ex::Expr) = ex.args[2:end]
-TermInterface.similarterm(x::Type{Expr}, head, args, symtype=nothing; metadata=nothing) = 
-    Expr(:call, head, args...)
-
 TermInterface.issym(s::Symbol) = true
 Base.nameof(s::Symbol) = s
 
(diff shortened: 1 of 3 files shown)


test/interface.jl
```diff
@@ -1,12 +1,6 @@
 using SymbolicUtils, Test
 using TermInterface
 
-TermInterface.istree(ex::Expr) = ex.head == :call
-TermInterface.operation(ex::Expr) = ex.args[1]
-TermInterface.arguments(ex::Expr) = ex.args[2:end]
-TermInterface.similarterm(x::Type{Expr}, head, args, symtype=nothing; metadata=nothing) = 
-    Expr(:call, head, args...)
-
```
shashi, line 9: How does this work now?

0x0f0f0f: https://github.com/JuliaSymbolics/TermInterface.jl/blob/master/src/expr.jl

shashi:
The symbolics test failure is a real one. Need to fix it.

0x0f0f0f:
> The symbolics test failure is a real one. Need to fix it.

Fixed in https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/368/commits/3df6bf194a4f1750574defe2c25ee18e2a4e1fc4

0x0f0f0f:
@shashi The buildkite test is failing saying that a function (that is defined!) is not defined. 
https://buildkite.com/julialang/symbolicutils-dot-jl/builds/161#ab7ec85c-0ae5-4bd6-ab38-d34b6fbac1e2/261-557

0x0f0f0f:
Superseded by #368

PR #365, opened 2021-09-14
Optimize Add/Mul ordering

YingboMa:


shashi:
maybe would be nice to have some motivating tests?

bolognam:
@YingboMa now I don't need this anymore :)

PR #373, opened 2021-09-22
simplifying Pow of Div + test

### LINKED ISSUE #372: simplify_fractions misses powers of fractions
simplify_fractions does not correctly unwrap a Div object from inside a Pow object.

`@syms x`
`term = (1/x)^2 * x^2`
`simplify_fractions(term) = (1/x)^2 * x^2`

Expanding does not help since 
`expand( (1/x)^2 ) = (1/x)^2`    (still a `Pow`, not a `Div`)

Might be good to either make `expand` unwrap the `Pow` , add similar functionality to `simplify_fractions` or to its subroutine `quick_cancel`

jkosata:
Fixes #372


src/polyform.jl
```diff
@@ -363,6 +364,10 @@ Has optimized processes for `Mul` and `Pow` terms.
 """
 quick_cancel(d::Div) = Div{symtype(d)}(quick_cancel(d.num, d.den)...)
 
+quick_cancel(x::Pow) = x.base isa Div ? Div(x.base.num^x.exp, x.base.den^x.exp) : x
+
+quick_cancel(x::Mul) = any(a->a isa Div, unsorted_arguments(x)) ? expand(x) : x
```
shashi, line 369: I think this won't be necessary if you used 

```
(x.base.num^x.exp) / (x.base.den^x.exp) : x
```
in the above function. `expand` is pretty slow, and we want to really really keep `quick_cancel` as fast as possible. :)

shashi: Actually you may need:

```
quick_cancel(x::Pow) = x.base isa Div ? quick_cancel((x.base.num^x.exp) / (x.base.den^x.exp)) : x
```

shashi (changes requested):

shashi:
Thanks for taking this on :) This is great!

src/polyform.jl
jkosata: I understand not wanting to use `expand` , but without it I always end up with
`simplify_fractions((1 / x) ^ 2 * x ^ 2)` ---> `(x^2)*(1 / (x^2))`

Oddly a similar manual input is automatically converted to a `Div`
`term = (x^2)*(1 / (x^2))` ---> `x^2 / x^2`
 `simplify_fractions(term)` ---> `1`

Which function is invoked to do this?

shashi:
@etiennedeg thanks for the bump.

Issue #344, opened 2021-08-24
Proposing an extension to @rule syntax (mirror julia syntax)


peterahrens:
Wouldn't it be nifty to match segment variables with `~a...`? I think it would be nice. It wouldn't be a difficult feature to add either, since we can detect splats in quoted expressions. The feature is non-breaking, since we can keep the old syntax. Furthermore, we need only make a change to `makepattern`. Something like this:

```
if expr.head === :call
    if expr.args[1] === :(~)
        #proceed normally...
elseif expr.head === :...
    if expr.args[1].args[1] === :(~)
        makesegment(expr.args[1].args[2].args[2], keys)
    end
else
#proceed normally...
```

0x0f0f0f:
I agree

0x0f0f0f:
I'm currently porting the SU syntax and rule language to Metatheory.jl, I'd redesign the SU syntax for rules but it's hard to make those breaking changes and we would have to PR all packages downstream. I would like to remove the `~` for pattern variables and use `...` for slot variables only, leaving the predicate syntax as is. Also, I would use `|>` for rules that evaluate the rhs (SU rules) and `=>` for rules that perform symbolic substitution only.
```
(~a*~b)*(~c...)-->~a*(~b*~c...)
```
is not really that readable, i'd propose
```
(a*b)*c...=>a*(b*c...)
```
 @peterahrens

peterahrens:
The `makepattern` macro relies on `~` to determine which symbols to turn into slots. Without it, we would need to declare which variables are slots or escape non-slot symbols. If I had a function `f` and a literal `a`, which of the following syntaxes would you prefer?

`@rule b c c * f(a, b) => g(c, b)`
`@rule c * $f($a, b) => g(c, b)`
`@rule c_ * f(a, b_) => g(c, b)` (This is the MacroTools.jl syntax)

The following option isn't purely syntactic; we would traverse the lhs to discover which nodes are typed as slots. The traversal isn't difficult, but it doesn't work well with the @capture macro (#342) because the @capture macro needs a list of variable names to inject into the environment at macro expansion time.
```
@slots b c
@rule c * f(a, b) => g(c, b)
```

I agree that the `~` syntax is a little messy, but I'm not sure the other options are better. I wish that `~` bound tighter. I was hoping to update my proposal for the capture macro with a better syntax "once I figured it out", but that didn't happen.

[...]

PR #368, opened 2021-09-16
[WIP] Move generic term rewriting features to Metatheory.jl

0x0f0f0f:
This is an attempt to unify the Symbolics ecosystem following a layered protocol approach, in order to maintain code as general as possible without sacrificing all the performance optimizations of SU, and in order to extend SU support to e-graph rewriting, non-evaluating symbolic-replacement rules (`-->` operator) and a new, drop-in, alternative friendly syntax:

- TermInterface is the glue package between all symbolics manipulation packages
- Metatheory.jl provides all the *generic* term rewriting utilities (Rule, ACRule, EGraphs, matchers, Rewriters,...) on any type that satisfies TermInterface.jl
- SymbolicUtils.jl provides the specifically optimized types and simplification rules for mathematical symbolic expressions (Add, Mul, Div, PolyForm), and code generation etc...

This follows the the last diagram of https://discourse.julialang.org/t/a-recap-of-the-state-of-the-symbolics-ecosystem/68022 and should in principle simplify the dependencies, CI fragmentation and versioning.

This needs https://github.com/JuliaSymbolics/Metatheory.jl/pull/77 to be merged and released first

The imported modules from Metatheory are
```
using Metatheory # generic term rewriting 
using Metatheory.Rules # RewriteRule, ACRule, DynamicRule
using Metatheory.Rewriters # Rewriter combinators
using Metatheory.SUSyntax # Use the SymbolicUtils rule language for defining rules.
```

If you want to try out the new syntax, replace `SUSyntax` with `NewSyntax`. In the new syntax, symbols in patterns (`x`) are slot variables, `x...` is used for segment variables, `=>` for substitution rules, `|>` for dynamic rules, `==` and `!=` for bidirectional e-graph specific rules.

PR #374, opened 2021-09-29
add @slots macro to improve readability of rules

peterahrens:
My proposed, backwards-compatible rule syntax extension discussed in #344. This PR allows users to write, e.g., a chain of rules like:

```
@slots x y z a b c Chain([
    (@rule x^2 + 2xy + y^2 -> (x + y)^2),
    (@rule x^ay^b -> (xy)^ay^(b-a)),
    (@rule +(x...) -> sum(x)),
])
```

We could write a single rule as

`@rule a b a + b -> add(a, b)`

or 

`@slots a b @rule a + b -> add(a, b)`

This is accomplished by making 4 changes:
1. `@capture` and `@rule` take an optional (prefixed) list of symbols to turn into slots, e.g. `@rule a b a + b -> add(a, b)`
2. `@slots syms... expr` adds the list of symbols `syms` to the beginning of each macrocall `@rule` `@capture` or `@slots` in the expression `expr`.
3. If a slot in a pattern would be the argument to a `...` expression, it is interpreted as a segment.
4. Matched variables are injected directly as variables into the environment of the rhs of an `@rule`.

This change is backwards-compatible, as it still recognizes the `~a` and `~~a` syntax. The only instance where this would be breaking is if the user expected to be able to use the variable `a` in the rhs when `a` is also a slot or segment in the lhs. We could avoid this by only injecting variables declared with the new syntax, but I think it's a rare case anyway.

This PR is missing docs and tests, but I am happy to add those if there is interest.

### DIFF
--- a/src/SymbolicUtils.jl
+++ b/src/SymbolicUtils.jl
@@ -35,7 +35,7 @@ include("rewriters.jl")
 using .Rewriters
 
 using Combinatorics: permutations, combinations
-export @rule, @acrule, RuleSet, @capture
+export @rule, @acrule, RuleSet, @capture, @slots
 
 # Rule type and @rule macro
 include("rule.jl")
--- a/src/rule.jl
+++ b/src/rule.jl
@@ -55,7 +55,7 @@ function makeslot(s::Expr, keys)
     :(Slot($(QuoteNode(name)), $(esc(s.args[2]))))
 end
 
-function makepattern(expr, keys)
+function makepattern(expr, keys, slots)
     if expr isa Expr
         if expr.head === :call
             if expr.args[1] === :(~)
@@ -67,15 +67,31 @@ function makepattern(expr, keys)
                     makeslot(expr.args[2], keys)
                 end
             else
-                :(term($(map(x->makepattern(x, keys), expr.args)...); type=Any))
+                :(term($(map(x->makepattern(x, keys, slots), expr.args)...); type=Any))
+            end
+        elseif expr.head === :...
+            if expr.args[1] === :(~)
+                if expr.args[2] isa Expr && expr.args[2].args[1] == :(~)
+                    # matches ~~x...
+                    makesegment(expr.args[2].args[2], keys)
+                else
+                    # matches ~x...
+                    makesegment(expr.args[2], keys)
+                end
+            elseif expr.args[1] in slots
+                makesegment(expr.args[1], keys)
+            else
+                Expr(:..., makepattern.(expr.args, (keys,), (slots,))...)
             end
         elseif expr.head === :ref
-            :(term(getindex, $(map(x->makepattern(x, keys), expr.args)...); type=Any))
+            :(term(getindex, $(map(x->makepattern(x, keys, slots), expr.args)...); type=Any))
         elseif expr.head === :$
             return esc(expr.args[1])
         else
-            Expr(expr.head, makepattern.(expr.args, (keys,))...)
+            Expr(expr.head, makepattern.(expr.args, (keys,), (slots,))...)
         end
+    elseif expr in slots
+        makeslot(expr, keys)
     else
         # treat as a literal
         return esc(expr)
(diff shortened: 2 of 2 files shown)


peterahrens:
For comparison, the first example in the current syntax looks like

```
Chain([
    (@rule (~x)^2 + 2(~x)*~y + (~y)^2 => (~x + ~y)^2),
    (@rule (~x)^~a * (~y)^~b => ((~x)*~y)^(~a) * (~y)^((~b)-(~a))),
    (@rule +(~~x) => sum(~~x)),
])
```

shashi:
Let's get some docs and merge this!

peterahrens:
I would also like to move forward with this PR.

@0x0f0f0f has expressed in the SymbolicManipulation slack that https://github.com/JuliaSymbolics/Metatheory.jl/pull/77 introduces a different `~`-less syntax into MetaTheory.jl. If I understand correctly, the new syntax referred to is described in the file named [NewSyntax.jl](https://github.com/JuliaSymbolics/Metatheory.jl/blob/d5bf09c4597aa4e67b951661929dbeba10c4d5f1/src/NewSyntax.jl). The new syntax assumes any symbol that isn't the function name in a function call to be a slot variable. As an example, for the expression `a + f(b)`, @0x0f0f0f's new syntax would turn `a` and `b` into slots, but leave `+` and `f` as coming from the environment.

Unfortunately, @0x0f0f0f's new syntax would be a breaking change. It is not backwards-compatible with the current SymbolicUtils syntax, which expects raw symbols to default to their values in the current environment. If @0x0f0f0f is alright with using `~` and the `@slots` syntax to specify slot variables instead, then I think these changes would be compatible with the other changes in https://github.com/JuliaSymbolics/Metatheory.jl/pull/77, which are largely different rule meanings for different kinds of arrows (e.g. `=>`, `|>` or `-->`).

MasonProtter:
I'm pretty against using that syntax in SymbolicUtils

peterahrens:
Please correct me if I'm wrong: After reading https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/368, I now see that this syntax doesn't need to play nice with MetaTheory's new syntax. This change is an extension to SU syntax, thus it would belong in MetaTheory's SUSyntax module. This change is orthogonal to the alternative, breaking, NewSyntax module being considered for MetaTheory.

0x0f0f0f:
> Please feel free to correct me if I'm wrong: After reading #368, I now see that this syntax doesn't need to play nice with MetaTheory's new syntax. This change is an extension to SU syntax, therefore corresponding changes would belong in MetaTheory's SUSyntax module. This change is orthogonal to the alternative, breaking, NewSyntax module being considered for MetaTheory.

Peter could you move this PR to Metatheory.jl/pull/77 please? It would avoid merging conflicts when #368 is done :)

peterahrens:
If merging PRs into SymbolicUtils is not an option at this time, perhaps I should wait for things to settle a bit before merging into Metatheory.jl instead?

On the other hand, if you were planning to unify the Metatheory and SymbolicUtils syntaxes in Metatheory.jl/pull/77, then of course I can merge this into that PR.

peterahrens:
It would help me if you could clarify what you mean by "unify the syntaxes." Are you referring to the solution that currently lives in MT/77 (separate submodules for two different syntaxes)? 

If you're referring to a solution with only one syntax for both SU and MT would this unified syntax be backwards-compatible with current SU syntax or with the current MT syntax?

peterahrens:
Thanks for clarifying! Since the syntaxes will be compatible, I think the right thing to do is to merge this PR into SymbolicUtils first, so that there are tests and a well-defined syntax to compare against when you're testing MT77. @0x0f0f0f or I can copy the code to MT77 (@0x0f0f0f, let me know if you'd like a PR to your PR). Regarding merge conflicts, @0x0f0f0f can simply delete the code this PR introduces when resolving conflicts in https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/368. That way, https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/368 and MT/77 can be simpler changes (closer to only moving code around).

0x0f0f0f:
> Thanks for clarifying! Since the syntaxes will be compatible, I think the right thing to do is to merge this PR into SymbolicUtils first, so that there are tests and a well-defined syntax to compare against when testing MT77. @0x0f0f0f or I can copy the code to MT77 (@0x0f0f0f, let me know if you'd like a PR to your PR). Regarding merge conflicts, @0x0f0f0f can simply delete the code this PR introduces when resolving conflicts in #368. That way, #368 and MT/77 can be made simpler (closer to only moving code around).

I just ported everything to MT77. I can delete in #368 later or we can close this and port everything that is left (docs and tests) to MT77 and then to #368.

PR #375, opened 2021-10-02
Fix printing of binary operators with 1 arg

MasonProtter:
Before this fix:
```julia
julia> SymbolicUtils.show_call(stdout, (-), 1)
1
```
After this fix:
```julia
julia> SymbolicUtils.show_call(stdout, (-), 1)
-(1)
```

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -499,7 +499,7 @@ end
 function show_call(io, f, args)
     fname = istree(f) ? Symbol(repr(f)) : nameof(f)
     binary = Base.isbinaryoperator(fname)
-    if binary
+    if binary && length(args) == 2
         for (i, t) in enumerate(args)
             i != 1 && print(io, " $fname ")
             print_arg(io, t, paren=true)


MasonProtter:
Hm, this seems to have somehow broken the printing of more complicated expressions for me, etc.
```julia
julia> 1 + -1*σ1
1 - (-((-σ1)^1))
```
:(

__________

```julia
using SymbolicUtils
using SymbolicUtils.Rewriters
using SymbolicUtils: Sym, Term, Symbolic, sym_isa
SymbolicUtils.show_simplified[] = false


struct Clifford end

@syms σ1::Clifford σ2::Clifford
@syms a b c

for op ∈ (:*, :+, :-, :^, :/)
    @eval begin
        Base.$op(x::Symbolic{Clifford}, y::Symbolic{Clifford}) = Term{Clifford}($op, [x, y])
        Base.$op(x, y::Symbolic{Clifford}) = Term{Clifford}($op, [x, y])
        Base.$op(x::Symbolic{Clifford}, y) = Term{Clifford}($op, [x, y])

        Base.$op(x::Number, y::Symbolic{Clifford}) = Term{Clifford}($op, [x, y])
        Base.$op(x::Symbolic{Clifford}, y::Number) = Term{Clifford}($op, [x, y])
    end
end
```

MasonProtter:
I think that the way that we handle the printing here is just bad. I don't understand why not just convert the `Term` to an `Expr` and then use `Base`'s printing.

shashi:
Seems to have gotten fixed over time.

PR #379, opened 2021-10-10
WIP: Preserve metadata in simplify_fractions

YingboMa:


YingboMa:
@shashi the test still fails. Do you have some clues?

PR #406, opened 2021-11-23
WIP: create_bindings in DestructuredArgs

shashi:
Allow avoiding creating let bindings if necessary.


shashi:
Hmm, those failures are numerical and deviate significantly outside of tolerance, I doubt this change is the cause, but if anyone thinks it's likely let me know. cc @ChrisRackauckas @YingboMa

shashi:
cc @KirillZubov

ChrisRackauckas:
Does it work on v1.6? I heard Flux has some issues with restructure/destructure on v1.7.

shashi:
it seems to work on 1.6 btw...

shashi:
We have the same failure on master without this change (on v1.7). I think the tests there need to be updated or `@test_broken`.

PR #409, opened 2021-12-03
ndims() and broadcast() definitions for generic symbolic types

gronniger:
This PR defines ndims() in terms of the type that a symbolic expression represents. That should be correct also for future/custom types.
Broadcast assumes all types are "scalar" until defined otherwise.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -157,4 +157,4 @@ Base.size(x::Symbolic{<:Number}) = ()
 Base.length(x::Symbolic{<:Number}) = 1
 Base.ndims(x::Symbolic{T}) where {T} = Base.ndims(T)
 Base.ndims(::Type{<:Symbolic{T}}) where {T} = Base.ndims(T)
-Base.broadcastable(x::Symbolic{T}) where {T} = Ref(x)
+Base.broadcastable(x::Symbolic{T}) where {T<:Number} = Ref(x)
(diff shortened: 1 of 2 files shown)


src/methods.jl
```diff
@@ -155,4 +155,6 @@ Base.literal_pow(::typeof(^), x::Symbolic{<:Number}, ::Val{p}) where {p} = Base.
 # Array-like operations
 Base.size(x::Symbolic{<:Number}) = ()
 Base.length(x::Symbolic{<:Number}) = 1
-Base.ndims(x::Symbolic{<:Number}) = 0
+Base.ndims(x::Symbolic{T}) where {T} = Base.ndims(T)
+Base.ndims(::Type{<:Symbolic{T}}) where {T} = Base.ndims(T)
+Base.broadcastable(x::Symbolic{T}) where {T} = Ref(x)
```
shashi, line 160: Maybe should restrict this to `T<:Number`?

gronniger: Actually, I think just `T` should be correct here as a reasonable default. Any more specific implementation could still be defined, but likely would look the same for any "scalar" type.
Otherwise, the test for the custom `SquareDummy` type would also fail.

shashi: I think it is correct only if Base had `Base.broadcastable(x::Any) = Ref(x)`. But it goes to some lengths to avoid defining it.
```julia
julia> @which Base.broadcastable("hi")
broadcastable(x::Union{Missing, Nothing, UndefInitializer, AbstractPattern, AbstractString, Function, Symbol, Pair, Val, Ptr, RoundingMode}) in Base.Broadcast at broadcast.jl:678
```

PR #413, opened 2021-12-25
Added ordering for types

gronniger:
This is required to make all tests in https://github.com/JuliaSymbolics/Symbolics.jl/pull/483 work.


shashi:
Would be nice to have a test

PR #414, opened 2022-01-02
Fix TRIG_EXP_RULES

ErikQQY:
I think **sin(x)^2-1** and **cos(x)^2-1** in ```TRIG_EXP_RULES``` is not right

Also I add @eqtest for that


shashi:
Thanks :)

PR #415, opened 2022-01-05
fix typos in Func toexpr docstring

guanyilun:
Spotted some typos in the docstring which confused me a little and fixed them here.


shashi:
Thanks!

PR #416, opened 2022-01-06
Stable hash and don't actually lower to Let

YingboMa:


src/code.jl
```diff
@@ -264,12 +264,12 @@ function toexpr(l::Let, st)
...
-    l.let_block ? Expr(:let,
-                       Expr(:block, bindings...),
-                       toexpr(l.body, st)) : Expr(:block,
-                                                  bindings...,
-                                                  toexpr(l.body, st))
+    ex = Expr(:block)
+    for p in dargs
+        push!(ex.args, toexpr(p, st))
+    end
+    push!(ex.args, toexpr(l.body, st))
```
shashi, line 271: The `l.let_block = false` condition does the same thing as this though? Can we just set it wherever you need this to happen?

YingboMa: What does it do?

shashi: removes the let block

PR #418, opened 2022-01-13
adds div and simplifications for div, mod and rem

rmsrosa:
I also added a bunch of tests, which highlight the simplifications added.

There are still some simplifications that can be included. I will work on those, but I think/hope this is a good start.

I haven't added anything to the documentation either. I can work on that later once I figure out how it is deployed.

Let me know if there is anything I should do differently. In particular regarding the way I added the rules to `default_simplifier`. I wanted to do minimal changes to that, but I suspect separating div/rem/mod from trig/exp simplifications might be best since they will certainly be less used.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -14,7 +14,7 @@ const monadic = [deg2rad, rad2deg, transpose, asind, log1p, acsch,
                  trigamma, invdigamma, polygamma, airyai, airyaiprime, airybi,
                  airybiprime, besselj0, besselj1, bessely0, bessely1]
 
-const diadic = [max, min, hypot, atan, mod, rem, copysign,
+const diadic = [max, min, hypot, atan, div, rem, mod, copysign,
                 besselj, bessely, besseli, besselk, hankelh1, hankelh2,
                 polygamma, beta, logbeta]
 const previously_declared_for = Set([])
--- a/src/simplify_rules.jl
+++ b/src/simplify_rules.jl
@@ -33,7 +33,6 @@ let
         @rule(*(~x) => ~x)
     ]
 
-
     POW_RULES = [
         @rule(^(*(~~x), ~y::_isinteger) => *(map(a->pow(a, ~y), ~~x)...))
         @rule((((~x)^(~p::_isinteger))^(~q::_isinteger)) => (~x)^((~p)*(~q)))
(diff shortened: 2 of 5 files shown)


rmsrosa:
Ok, I made my last commit of the series. I added the missing simplifications and adjusted testing. It is good to go.

src/simplify_rules.jl
```diff
@@ -132,10 +158,11 @@ let
     global serial_expand_simplifier
 
     function default_simplifier(; kw...)
-        IfElse(has_trig_exp,
+        IfElse(t -> (has_trig_exp(t) || has_div_rem_mod(t)),
```
shashi, line 161: I think this needs a different branch not just `||`.

rmsrosa: Do you mean one branch for `&&` and two others for either one?

shashi: No I mean that there should be another IfElse node..

`IfElse(has_div_rem_mod, Chain((divremmod_simplifier(), simplifier), simplifier)`

where simplifier is the object from this line as it was before.

shashi: It seems weird that we would want to apply the trig simplifier if there is div mod and vice versa.

shashi (changes requested):

src/simplify_rules.jl
rmsrosa: Oh, I thought we needed both chained at this point to simplify expressions with both terms, but apparently I misunderstood it. Ok, I just made a new commit with different branches and the tests are good.

src/simplify_rules.jl
```diff
@@ -133,18 +159,35 @@ let
...
+                ),            
+                Postwalk(
+                    Chain(
+                        (If(x->symtype(x) <: Number, number_simplifier()),
+                        If(x->symtype(x) <: Bool, bool_simplifier()))
+                    )
+                    ; kw...
+                )
+            )
+        )
```
shashi, line 190: This does not work when both predicates are true. How about something like:

```julia
chained(xs...) =  Postwalk(
                IfElse(
                    x->symtype(x) <: Number,
                    Chain(xs),
                    If(x->symtype(x) <: Bool, bool_simplifier())
                )
                ; kw...
            )
IfElse(has_trig_exp,
          IfElse(has_div_rem_mod,
               chained(number_simplifier(),  trig_exp_simplifier(), div_rem_mod_simplifier()),
               chained(number_simplifier(),  trig_exp_simplifier())),
          chained(number_simplifier()))
```

rmsrosa: This should fail if there is only div/rem/mod. But the use of `chained` should help. I will try something.

rmsrosa: It turns out the solution is much simpler:
```julia
    default_simplifier(; kw...) = Postwalk(
        Chain(
            (
                If(x->symtype(x) <: Number, number_simplifier()),
                If(has_trig_exp, trig_exp_simplifier()),
                If(has_div_rem_mod, div_rem_mod_simplifier()),
                If(x->symtype(x) <: Bool, bool_simplifier())
            )
        )
        ; kw...
    )
```

All test are good.

rmsrosa: It even seemed to have improved the performance, although I am not sure what I got with a single comparison is representative.

src/methods.jl
```diff
@@ -14,7 +14,7 @@ const monadic = [deg2rad, rad2deg, transpose, asind, log1p, acsch,
                  trigamma, invdigamma, polygamma, airyai, airyaiprime, airybi,
                  airybiprime, besselj0, besselj1, bessely0, bessely1]
 
-const diadic = [max, min, hypot, atan, mod, rem, copysign,
+const diadic = [max, min, hypot, atan, div, rem, mod, copysign,
```
YingboMa, line 17: I hope we can directly simplify `div` on construction.

rmsrosa: How can I do that?

Is the same planned for trig/exp?

YingboMa: Check the `src/types.jl` file for Div overloads.

YingboMa (approved):

src/simplify_rules.jl
```diff
@@ -132,20 +158,17 @@ let
...
-                                  number_simplifier()),
-                               If(x->symtype(x) <: Bool,
-                                  bool_simplifier())))
-                        ; kw...))
-    end
+    default_simplifier(; kw...) = Postwalk(
+        Chain(
+            (
+                If(x->symtype(x) <: Number, number_simplifier()),
+                If(has_trig_exp, trig_exp_simplifier()),
```
YingboMa, line 165: `has_*` could be expensive. Maybe we should have one combined `has_` function that returns `has_trig_exp, has_div_rem_mod` so there's only one traversal. This can be done in a later PR.

rmsrosa:
Hmm, is there anything missing from my part? It still shows a change request but I believe I addressed that, already. I thought this was good to go. Let me know if I am missing something.

YingboMa:
@shashi could you take a look?

PR #421, opened 2022-01-17
Call quick cancel by default

YingboMa:


YingboMa:
With this branch
```julia
julia> simplify_fractions.(simplify.(inv(A), expand=true)) * det(A)
2×2 Matrix{Num}:
  a[2, 2]  -a[1, 2]
 -a[2, 1]   a[1, 1]
```
On master:
```julia
julia> simplify_fractions.(simplify.(inv(A), expand=true)) * det(A)
2×2 Matrix{Num}:
  a[2, 2]  -a[1, 2]
 -a[2, 1]   a[1, 1]

julia> simplify_fractions.(simplify.(inv(A), expand=true)) * det(A)
2×2 Matrix{Num}:
  ((a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])*a[2, 2]) / (a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])  …  (-(a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])*a[1, 2]) / (a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])
 (-(a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])*a[2, 1]) / (a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])      ((a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])*a[1, 1]) / (a[1, 1]*a[2, 2] - a[1, 2]*a[2, 1])
```

shashi:
I'm not looking forward to issues about how this is wrong.. Besides the right fix for the problem you quote is in #422 

If we do this, I would also like to have `SafeReal` type in the same release.

shashi:
Done in #425

Issue #397, opened 2021-11-12
simplify_fractions may have some problems?


liushang0322:
this maybe right:
@syms G1 G2 G3 G4 H1 H2 H3
Q = zeros(Num,5,5)
Q[1,5] = -H3
Q[2,1] =  G1
Q[2,4] = -H2
Q[3,2] =  G2
Q[3,5] =  H1
Q[4,3] =  G3
Q[5,4] =  G4

P=[1 0 0 0 0]'
G=(I(size(Q,1))-Q)\P
g = G[end]
g =  expand(g)
#= g = simplify_fractions(g) =#
op = operation(g)
g1,g2 = arguments(g)
g2 = PolyForm(g2)
g2 = simplify_fractions(g2)
g = g1//g2

but  this is wrong?

@syms G1 G2 G3 G4 H1 H2 H3
Q = zeros(Num,5,5)
Q[1,5] = -H3
Q[2,1] =  G1
Q[2,4] = -H2
Q[3,2] =  G2
Q[3,5] =  H1
Q[4,3] =  G3
Q[5,4] =  G4

P=[1 0 0 0 0]'
G=(I(size(Q,1))-Q)\P
g = G[end]
g =  expand(g)
g = simplify_fractions(g)

Issue #399, opened 2021-11-15
simplify simplify_fractions maybe too weak for polyform


liushang0322:
Another problem is that when the expression has many operations, it is difficult to simplify the expression to something like matlab, especially when numden is used in matlab, the arguments seem to be very weak, and only a single operator can be simply taken out. Operands, but cannot be decomposed well, especially when the final form is polynomial/polynomial.

shashi:
I could not figure out what the problem is from your description. Could you please post an example?

liushang0322:
https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/397

liushang0322:
> could not figure out what the problem is from your description. Could you please post an example?

could you please have a look ? thank you too much! https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/397

PR #422, opened 2022-01-17
Fixes to simplify_fractions, call it in simplify by default

shashi:
Fixes https://github.com/JuliaSymbolics/Symbolics.jl/issues/507, #397, #399, https://github.com/SciML/ModelingToolkit.jl/issues/1392


shashi:
Also fixes

#397 #399

PR #442, opened 2022-04-29
allow kws in `(f::Symbolic{<:FnType})(args...)` 

anandijain:
MTK `convert_system` uses `operation(s)(new_iv)`, which loses the metadata.


YingboMa (changes requested):
`Any[]` is important here/

anandijain:
`similarterm` resolved the issue

PR #450, opened 2022-06-30
Simplify printing of Term(identity, x)

moble:
It can be helpful to encode a quantity `q` that should not be evaluated as `Term(identity, q)`.  This is the idea behind JuliaSymbolics/Symbolics.jl#638, which allows irrationals like `π` to be treated symbolically, rather than (typically) as `Float64` in symbolic expressions.  But that means that `identity(π)` appears everywhere we would normally just write `π`.  This PR simply adds another case to `show_term`, to check if the function is `identity`, and just shows the argument instead.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -845,6 +845,8 @@ function show_term(io::IO, t)
         show_pow(io, args)
     elseif f === (getindex)
         show_ref(io, f, args)
+    elseif f === (identity)
+        show(io, args[1])
     else
         show_call(io, f, args)
     end


moble:
@shashi @YingboMa Given that https://github.com/JuliaSymbolics/Symbolics.jl/pull/638 has been merged, I think this is more relevant now.  Any thoughts?

shashi:
Ahh. Well I don't know how I feel about this haha. Can we take this branch only if `args[1]` is not `issym(..)` or `istree(..)` that way it should only do it for constants. At the very least I agree we should do it for Irrationals. @YingboMa

YingboMa:
Yeah, I agree. We should check that `args[1]` is not symbolic.

src/types.jl
```diff
@@ -845,6 +845,8 @@ function show_term(io::IO, t)
         show_pow(io, args)
     elseif f === (getindex)
         show_ref(io, f, args)
+    elseif f === (identity)
```
moble, line 848: So @shashi and @YingboMa, you're suggesting this?

```suggestion
    elseif f === (identity) && !issym(args[1])
```

If so, I have no problem with this.

shashi: Also !istree(args[1]]

shashi: There's a syntax error. Could you also add a little test?

moble: Will do.

moble:
@shashi Tests pass locally.  I added the new test at the bottom of `test/code.jl`; let me know if you want it somewhere else.

shashi:
You can just use `repr` to get a string... but looks good.

Issue #437, opened 2022-03-08
Should 1^x == 1?


pagnani:
Following the [discussion](https://discourse.julialang.org/t/should-1-x-1/77529/5?u=andrea_pagnani)  on Discourse on the same issue:
do we want to enforce a rule to make 1^x -> 1?

There is a concern for x = ±∞ for which, I believe, the substitution although mathematically disputable should be ok.

As variables are strongly typed, a concern is what happens with `one(ComplexF*)`?  However, it seems  that 

`SymbolicUtils._isone(one(ComplexF64))` is true.

BTW, I tried to add to`POW_RULES` something like

```
@rule(^(~x::_isone, ~z) => 1)
```

but it does not like it (stacktrace below). I do not understand how one can add more rules ....

```
ERROR: LoadError: ArgumentError: number of columns of each array must match (got (1, 1, 2, 1, 2, 1, 2, 1))
Stacktrace:
  [1] _typed_vcat(#unused#::Type{Metatheory.Rules.AbstractRule}, A::NTuple{8, Matrix{Metatheory.Rules.AbstractRule}})
    @ Base ./abstractarray.jl:1605
  [2] typed_vcat(::Type{Metatheory.Rules.AbstractRule}, ::Matrix{Metatheory.Rules.AbstractRule}, ::Matrix{Metatheory.Rules.AbstractRule}, ::Matrix{Metatheory.Rules.AbstractRule}, ::Vararg{Matrix{Metatheory.Rules.AbstractRule}})
    @ Base ./abstractarray.jl:1619
  [3] typed_hvcat(::Type{Metatheory.Rules.AbstractRule}, ::NTuple{8, Int64}, ::Metatheory.Rules.DynamicRule, ::Vararg{Any})
    @ Base ./abstractarray.jl:2041
  [4] hvcat(::NTuple{8, Int64}, ::Metatheory.Rules.DynamicRule, ::Vararg{Any})
    @ Base ./abstractarray.jl:2020
  [5] top-level scope
    @ ~/FORKS/SymbolicUtils.jl/src/simplify_rules.jl:4
  [6] include(mod::Module, _path::String)
    @ Base ./Base.jl:418
  [7] include(x::String)
    @ SymbolicUtils ~/FORKS/SymbolicUtils.jl/src/SymbolicUtils.jl:4
  [8] top-level scope
    @ ~/FORKS/SymbolicUtils.jl/src/SymbolicUtils.jl:55
  [9] include
[...]

PR #469, opened 2022-09-05
added rule for 1^x == 1 in POW_RULES

pagnani:
- Brought this up on [discourse](https://discourse.julialang.org/t/should-1-x-1/77529), and relative to issue [437](https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/437#issue-1162808282).
- It basically introduces a new simplification rule that sets, as for the title, the rule 1^x == 1, which should be ok for all x.
- Mathematica enforces it by default, and it seems a good idea to have it here.

### DIFF
--- a/src/simplify_rules.jl
+++ b/src/simplify_rules.jl
@@ -40,6 +40,7 @@ let
         @rule(^(~x, ~z::_iszero) => 1)
         @rule(^(~x, ~z::_isone) => ~x)
         @rule(inv(~x) => 1/(~x))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

YingboMa:
The failure is real.

pagnani:
Yeah, there was a typo in the test. Test now pass locally ...

pagnani:
Is there anything more to do here?

PR #475, opened 2022-10-06
add `create_array` for `ReshapedArray`

baggepinnen:


baggepinnen:
@YingboMa can we have this merged?

YingboMa (commented):
Could you add a test?

Issue #462, opened 2022-08-16
`similarterm` inconsistent with `Symbolic` arithmetic operations


bowenszhu:
`similarterm` hard-code the resulting `Symbolic` subtype according to the input operation `f`. For example, when `f` is `*`, `similarterm` calls `Mul` no matter what the input `args` are.

https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/97a8a86227f5eb824c3c28e5f2094d8168b5fa57/src/types.jl#L529-L549

However, the result's type is not necessarily `Mul` for the `*` operations defined in `src/types.jl`. 

https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/97a8a86227f5eb824c3c28e5f2094d8168b5fa57/src/types.jl#L1121-L1162

See the following test case.

```julia
julia> using SymbolicUtils

julia> @syms x y;

julia> s = x + y;

julia> m = -1 * s;

julia> typeof(m)
SymbolicUtils.Add{Number, Int64, Dict{SymbolicUtils.Sym{Number, Nothing}, Int64}, Nothing}

julia> m.coeff
0

julia> m.dict
Dict{SymbolicUtils.Sym{Number, Nothing}, Int64} with 2 entries:
  y => -1
  x => -1
```

The result of the normal `*` operation is `Add(coeff = 0, dict = Dict(x => -1, y => -1)`.

```julia
julia> using SymbolicUtils

julia> @syms x y;

julia> s = x + y;

julia> f = *;

julia> args = (-1, s);

julia> res = similarterm(SymbolicUtils.Mul, f, args, Real)
-(x + y)

julia> typeof(res)
SymbolicUtils.Mul{Real, Int64, Dict{Any, Number}, Nothing}

julia> res.coeff
-1

julia> res.dict
Dict{Any, Number} with 1 entry:
  x + y => 1
```

The result of `similarterm` is `Mul(coeff = -1, dict = Dict(Add(coeff = 0, dict = Dict(x => 1, y => 1)) => 1)`.

shashi:
Nice issue! I wonder if just calling `*` in `similarterm` would be the way to solve this. Do you have any ideas?

bowenszhu:
I guess so. The set of `Symbolic` arithmetic operations are well defined. It seems unnecessary to manually define new rules for the resultant types and data structures in `similarterm`.

PR #480, opened 2022-10-30
Call `BasicSymbolic` arithmetic operations to construct `similarterm`

bowenszhu:
# resolve https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/462
# resolve https://github.com/JuliaSymbolics/Symbolics.jl/issues/704

In the docstring of `similarterm`, it says that the return type may be different than `t`.

https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/6964c768e76dcbc48ee7b1d58fae481d46caa5a0/src/types.jl#L509-L514

However, the type is directly hard-coded according to `f` in the code.

https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/6964c768e76dcbc48ee7b1d58fae481d46caa5a0/src/types.jl#L530-L550

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -535,15 +535,12 @@ function TermInterface.similarterm(t::Type{<:BasicSymbolic{<:Number}}, f, args, 
     if T === nothing
         T = _promote_symtype(f, args)
     end
-    if f === (+)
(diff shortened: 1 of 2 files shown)


shashi:
could you review the parts you wrote in #482

PR #482, opened 2022-10-31
[wip] Unityper updates

shashi:
fixes `isequal` and adds missing `istree` for now.


shashi:
and https://github.com/JuliaSymbolics/Symbolics.jl/pull/549

src/types.jl
```diff
@@ -190,7 +190,7 @@ isdiv(x)  = isa_SymType(Val(:Div), x)
 Base.isequal(::Symbolic, x) = false
 Base.isequal(x, ::Symbolic) = false
 Base.isequal(::Symbolic, ::Symbolic) = false
-
+coeff_isequal(a, b) = isequal(a, b) || ((a isa AbstractFloat && b isa AbstractFloat) && (a==b))
```
bowenszhu, line 193:
```suggestion
coeff_isequal(a, b) = isequal(a, b) || ((a isa AbstractFloat || b isa AbstractFloat) && (a==b))
```
From the commit https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/972e675519c5a6a35763e6c657e49f73c0d3115a message, this seems to be used for dealing with the comparison of `Float64` `-0.0` and `Int` `0`. 
So maybe the condition here should be changed from both being floating-point to either one being floating-point?

YingboMa: Yeah, this makes sense.

shashi:
Also returns all the code from Metatheory.

src/interface.jl
```diff
@@ -0,0 +1,82 @@
+"""
+  istree(x)
```
bowenszhu, line 2: A 4-space indent is needed so that the signature is printed as Julia code in the documentation.
The same applies to other functions in this file.

bowenszhu, line 2:
```suggestion
    istree(x)
```
A 4-space indent is needed so that the signature is printed as Julia code in the documentation.
The same applies to other functions in this file.

PR #488, opened 2022-11-15
switch to Documenter

ArnoStrouwen:
Does most of the work required to switch to Documenter.
Switching to Documenter, means we can include this package in docs.sciml.ai .

Some docstrings are still missing, and need to figure out how to make the relative links work again.


ChrisRackauckas:
Thanks! Yes, valuable upgrade. @shashi make note that we're doing this so more people can easily help maintain the docs.

PR #508, opened 2023-02-08
Add upgrade guide

YingboMa:

PR #509, opened 2023-02-08
Fix documentation build

shashi:


YingboMa:
Good to merge?

ArnoStrouwen:
Don't merge this, since we added Documenter. If you add both, they will start overwriting each other.

shashi:
You need to help get #508 published.

ArnoStrouwen:
https://symbolicutils.juliasymbolics.org/dev/

> But where is the documenter output ?

PR #510, opened 2023-02-09
activate Documenter

ArnoStrouwen:
This is what still needs to be done to activate Documenter.
It was not done yet because it will trash what is currently in the gh-pages branch.


YingboMa:
What's the link for the docs?

ArnoStrouwen:
Currently only https://symbolicutils.juliasymbolics.org/dev/ works.
A new stable release is needed, and then it can be added to docs.sciml.ai.

PR #516, opened 2023-03-07
Fix bug preventing printing of callable structs

AayushSabharwal:
See https://github.com/JuliaSymbolics/Symbolics.jl/pull/846

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -777,7 +777,14 @@ function show_ref(io, f, args)
 end
 
 function show_call(io, f, args)
-    fname = istree(f) ? Symbol(repr(f)) : nameof(f)
+    fname = if istree(f)
+        Symbol(repr(f))
+    elseif f isa Function
+        nameof(f)
+    else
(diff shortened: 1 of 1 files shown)


AayushSabharwal:
It seems the only way to test this is to call `show_call` directly?

shashi:
You can use `repr`..

src/types.jl
```diff
@@ -777,7 +777,14 @@ function show_ref(io, f, args)
...
 
 function show_call(io, f, args)
-    fname = istree(f) ? Symbol(repr(f)) : nameof(f)
+    fname = if istree(f)
+        Symbol(repr(f))
+    elseif f isa Function
+        nameof(f)
+    else
+        # Dummy for callable structs
+        Symbol()
```
shashi, line 786: How does this actually show the name of the struct type? This should be also `Symbol(repr(f))` or something that gets the type name as a symbol in my opinion.

PR #521, opened 2023-04-18
remove piracy on isequal

vtjnash:
Base expects `isequal` to fallback to `==`. Changing that forces some unnecessary method invalidations, since the new and old definitions seem to be the same.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -185,8 +185,6 @@ isdiv(x)  = isa_SymType(Val(:Div), x)
 ### Base interface
 ###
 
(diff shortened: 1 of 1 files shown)


YingboMa:
We overload `==` to return `Symbolic` though.

shashi:
> This returns false, which is clearly not Symbolic?

This is `isequal`... `==` returns `Symbolic{Bool}` so we cannot fallback to it.

shashi:
> They seem like an ambiguity trap

Not been the case in practice, because the default `isequal` is defined using the common symbolic expression interface we use.

PR #525, opened 2023-06-14
Fix for 32bit Julia

bradcarman:


src/types.jl
```diff
@@ -223,22 +223,28 @@ Base.nameof(s::BasicSymbolic) = issym(s) ? s.name : error("None Sym BasicSymboli
 
 ## This is much faster than hash of an array of Any
 hashvec(xs, z) = foldr(hash, xs, init=z)
+
+const csym = typemax(UInt)
```
YingboMa, line 227: It should be something like `const SYM_SALT = 0x4de7d7c66d41da43 % UInt`

bradcarman: But this is a UInt64.  hash(::Symbol, ::UInt64) isn't supported by 32bit Julia (as best I can tell)

YingboMa: `0x4de7d7c66d41da43 % UInt` is UInt

bradcarman: In Julia 32bit, this is what I get...

```julia
julia> const SYM_SALT = 0x4de7d7c66d41da43
0x4de7d7c66d41da43

julia> typeof(SYM_SALT)
UInt64

julia> hash(:symbol, SYM_SALT)
ERROR: MethodError: no method matching hash(::Symbol, ::UInt64)
```

YingboMa: That's not `0x4de7d7c66d41da43 % UInt` though.

bradcarman: OK, got it, sorry I kept seeing something different 😅 .  Fixed now.

PR #530, opened 2023-06-26
Add x + conj(x) rule

amilsted:
Not sure if this is the right thing to do, but a PR seemed like the easiest way to bring this up.


shashi:
Fine with this, could you add a test or two?

amilsted:
> Fine with this, could you add a test or two?

Done. Also added a couple more rules (correct ones this time 😅).

MilesCranmer:
Just curious @shashi is there an `::_iscomplex`?

shashi:
@MilesCranmer No I don't think so, but it could be added.

amilsted:
Looks like the tests I added are passing now. I guess the segfault is a more general issue?

MilesCranmer:
@shashi The reason I asked about the `::_iscomplex` is that it looks like adding this rule would hurt the performance of simplification by about 20% across the board (https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/530#issuecomment-1608440859) even though I think some of those tests are real numbers only. So maybe having that `_iscomplex` check would reduce the impact of this on non-complex simplification problems?

PR #535, opened 2023-07-26
Use NaNMath lowering by default

YingboMa:
@ArnoStrouwen


isaacsas:
Does this have performance implications?

YingboMa:
I don't think there's any.

PR #536, opened 2023-07-31
Register NaNMath functions

YingboMa:


YingboMa:
The test failure seems unrelated.

PR #537, opened 2023-08-02
equality saturation

shashi:
simple, performance conscious port of the egg algorithm. Simplify works with it.


shashi:
**Tasks**:

- [ ] Only rules of depth 1 are working right now...
- [ ] acrules will not explore different order of applications, for example 2+1+5 will always be 3+5 then 8, never 2+6 then 8
- [ ] need some way to back off and break for infinite loops https://github.com/egraphs-good/egg/discussions/60 && https://julialang.slack.com/archives/C8R5HJMSN/p1683499144970229

Issue #546, opened 2023-08-30
Metadata is lost on construction of Mul


shashi:
(and potentially other constructions)

```julia
julia> SymbolicUtils.inspect(y)
1 POW metadata=(Sign => Positive,)
2 ├─ SYM(x) metadata=(Sign => AnySign,)
3 └─ 2

julia> SymbolicUtils.inspect(2y)
1 MUL(scalar = 2, powers = (x => 2,))
2 ├─ 2
3 └─ POW
4    ├─ SYM(x) metadata=(Sign => AnySign,)
5    └─ 2

```
when `x` is `LiteralReal`, thanks to #429 
```julia
julia> SymbolicUtils.inspect(2y)
1 TERM{*}
2 ├─ 2
3 └─ TERM{^} metadata=(Sign => Positive,)
4    ├─ SYM(x) metadata=(Sign => AnySign,)
5    └─ 2

```

The correct solution is to not canonicalize arguments with metadata.

shashi:
fixed by #547

PR #547, opened 2023-08-31
prevent canonicalization where arguments have metadata

shashi:
#546


shashi:
Oh wow this means that nothing in Symbolics will be canonicalized because of the `VariableName` metadata we attach in `@variables` -- @YingboMa are we able to get rid of that yet?

shashi:
Made it less conservative, it should continue to work 👍

YingboMa:
Unfortunately we still need the variable metadata.

PR #548, opened 2023-08-31
WIP: Various tweaks to the Rewriters

shashi:
- adds a `~MATCH` slot which holds the whole expression
- Adds a way for Chain to stop after finding a matching rule-useful when you want to chain rules by priority
- fix metadata propagation in Walk
- inspect & pluck (#538) is included here so this PR depends on that one.


src/rewriters.jl
```diff
@@ -195,8 +201,9 @@ function (p::Walk{ord, C, F, false})(x) where {ord, C, F}
         if ord === :pre
             x = p.rw(x)
         end
-        if iscall(x)
-            x = p.similarterm(x, operation(x), map(PassThrough(p), unsorted_arguments(x)))
+        if istree(x)
```
Vaibhavdixit02, line 204: This looks incorrect right?

shashi: yes

PR #557, opened 2023-10-13
NaNMath.pow fix

contradict:
NaNMath.pow needs a dispatch to prevent stack overflow with symbolic arguments.


contradict:
I created separate PR's to fix unrelated errors in Documentation (#558) and DataDrivenDiffEq (#559).

I have no idea how to fix the fuzz test error but it doesn't seem related.

I'm looking at the ModelingToolkit failure, but it doesn't look related.

PR #558, opened 2023-10-13
Update documentation generator

contradict:
Strict is now the default, need to specify exceptiopns instead.


bowenszhu (approved):
Yes. The `strict` keyword argument to `makedocs` was removed and replaced with `warnonly` in Documenter.jl v1.0.0 in September 2023.

https://github.com/JuliaDocs/Documenter.jl/blob/027544a3bfb02f70abb8430ece508edaa35873b5/CHANGELOG.md?plain=1#L98-L113

PR #559, opened 2023-10-13
Fix DataDrivenDiffEq.jl downstream test

contradict:
"Standard" group no longer exists in DataDrivenDiffEq.jl, "Core" looks like a good replacement.


bowenszhu (approved):
Indeed there is no `Standard` test group.

https://github.com/SciML/DataDrivenDiffEq.jl/blob/3ece13ddc126830ebdf4fd0e14ed9a820f07a1df/test/runtests.jl

PR #560, opened 2023-11-01
Add more trig identities

derikk:
Now we can have `simplify(1 - sin(x)^2) == cos(x)^2` instead of just `simplify(sin(x)^2 - 1) == -(cos(x)^2)`. 

Also fixed `@acrule(tan(~x)^2 + -1*sec(~x)^2 => one(~x))` by adding a minus sign.


oameye:
why is this kept from merging?

PR #566, opened 2023-12-11
refactor: add symbolic_type trait impl from SII

AayushSabharwal:


bowenszhu:
Hi @AayushSabharwal,
Could you provide some context about the purpose of this code?

AayushSabharwal:
SymbolicIndexingInterface.jl (SII) is a package that implements symbolic indexing for the SciML ecosystem. It has a trait to identify symbolic variables (since they don't have a common subtype) which is implemented here.

PR #585, opened 2024-03-11
Simplify the construction of `x / 1` and `x / -1`

YingboMa:


YingboMa:
The test failure is unrelated.

PR #584, opened 2024-03-10
New TermInterface

0x0f0f0f:

PR #587, opened 2024-03-14
Get tests to pass in new TermInterface

nmheim:
@0x0f0f0f here are my fixes for #584.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -136,6 +136,7 @@ function arguments(x::BasicSymbolic)
     end
     return args
 end
+children(x::BasicSymbolic) = arguments(x)
 function unsorted_arguments(x::BasicSymbolic)
     @compactified x::BasicSymbolic begin
         Term => return x.arguments
--- a/src/types.jl
+++ b/src/types.jl
@@ -816,7 +816,7 @@ function show_term(io::IO, t)
         show_pow(io, args)
     elseif f === (getindex)
         show_ref(io, f, args)
-    elseif f === (identity) && !issym(args[1]) && iscall(args[1])
+    elseif f === (identity) && !issym(args[1]) && !iscall(args[1])
         show(io, args[1])
     else
         show_call(io, f, args)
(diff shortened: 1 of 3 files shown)


src/types.jl
```diff
@@ -136,6 +136,7 @@ function arguments(x::BasicSymbolic)
     end
     return args
 end
+children(x::BasicSymbolic) = arguments(x)
```
nmheim, line 139: this could be fishy

src/utils.jl
```diff
@@ -221,7 +221,7 @@ macro matchable(expr)
         $expr
         SymbolicUtils.iscall(::$name) = true
         SymbolicUtils.operation(::$name) = $name
-        SymbolicUtils.arguments(x::$name) = getfield.((x,), ($(QuoteNode.(fields)...),))
+        SymbolicUtils.children(x::$name) = getfield.((x,), ($(QuoteNode.(fields)...),))
```
nmheim, line 224: as well as this

src/types.jl
```diff
@@ -813,7 +816,7 @@ function show_term(io::IO, t)
         show_pow(io, args)
     elseif f === (getindex)
         show_ref(io, f, args)
-    elseif f === (identity) && !issym(args[1]) && iscall(args[1])
+    elseif f === (identity) && !issym(args[1]) && !iscall(args[1])
```
0x0f0f0f, line 819: GOOD CATCH

src/types.jl
nmheim: this was what I was thinking as well, but then stumbled over `arguments` calling `unsorted_arguments` which is part of TI, so I kept it the current way. should we rename `unsorted_arguments` as well then? 🤔

0x0f0f0f: Ok, sounds good for now. Maybe we should file an issue to TI directly. cc @shashi

src/utils.jl
nmheim: ahhh, yes, that makes sense. missed the `iscall(::$name) = true` 👍

0x0f0f0f: Also missing `head`

src/types.jl
shashi: Shouldnt children also include the operation at the beginning?

PR #596, opened 2024-05-14
revert TermInterface integration

shashi:
- Fix https://github.com/JuliaSymbolics/Symbolics.jl/issues/1141
- Fix https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/595


shashi:
Can revert this revert and release a major version.

PR #598, opened 2024-05-17
bring back TermInterface

shashi:


src/types.jl
```diff
@@ -528,28 +535,12 @@ end
...
-
-Create a term that is similar in type to `t`. Extending this function allows packages
-using their own expression types with SymbolicUtils to define how new terms should
-be created. Note that `similarterm` may return an object that has a
-different type than `t`, because `f` also influences the result.
-
-## Arguments
+function TermInterface.maketerm(::Type{<:BasicSymbolic}, head, args, type, metadata)
+    basicsymbolic(first(args), args[2:end], type, metadata)
+end
```
0x0f0f0f, line 540: Please let's use the new one directly

0x0f0f0f:
```
function TermInterface.maketerm(::Type{<:BasicSymbolic}, op, args, type=nothing, metadata=nothing)
    basicsymbolic(op, args, type, metadata)
end
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
 authors = ["Shashi Gowda"]
-version = "1.7.0"
+version = "1.8.0"
```
0x0f0f0f, line 4: Version 2.0?

src/types.jl
nsajko: xref https://github.com/JuliaSymbolics/TermInterface.jl/issues/33

Issue #601, opened 2024-05-23
Casts from float64, int64, to symbolic types are missing


orebas:
I have any array of equations created by some other symbolic code, and it breaks when I try to substitute the value 1.0 for the rhs of an equation.

I just want a representation of the value 1.0, and I can't seem to create it in the obvious way (see below).

```
t = SymbolicUtils.BasicSymbolic{Real}(1.0)
ERROR: MethodError: no method matching SymbolicUtils.BasicSymbolic{Real}(::Float64)

Closest candidates are:
  (::Type{SymbolicUtils.BasicSymbolic{T}} where T)(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any)
   @ SymbolicUtils ~/.julia/packages/SymbolicUtils/qyMYa/src/types.jl:27

Stacktrace:
 [1] top-level scope
   @ REPL[18]:1
```

orebas:
Here's a very short MWE, which is similar to what I'm trying to do:
```
using ModelingToolkit
@variables x,y,z
 eqvec = [x ~ z, y~x^2+z]
rhs_vec = [e.rhs for e in eqvec]
#rhs_vec[1] = x
#rhs_vec[1] = x/x
rhs_vec[1] = 1.0
```

bowenszhu:
Please use 
```
Symbolics.wrap(SymbolicUtils.Term(identity, [1.0]))
```

Sorry. This is a known issue, and we still haven't provided a simpler constructor.
- https://github.com/JuliaSymbolics/Symbolics.jl/issues/853

bowenszhu:
Please use 
```
SymbolicUtils.Term{Real}(identity, [1.0])
```

Sorry. This is a known issue, and we still haven't provided a simpler way.
- https://github.com/JuliaSymbolics/Symbolics.jl/issues/853

orebas:
Actually, I still get this error:
ERROR: MethodError: Cannot `convert` an object of type Num to an object of
type SymbolicUtils.BasicSymbolic{Real}

the "typeof" the expression you recommended seems to be "Num"

bowenszhu:
Please remove `Symbolics.wrap` if you used it.

PR #602, opened 2024-05-25
Add `Symbolic` number constructor and conversion

bowenszhu:
- fix https://github.com/JuliaSymbolics/Symbolics.jl/issues/853
- fix https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/601

This pull request adds a constructor and a conversion function for `Symbolic` objects representing numbers.

#### Key changes:

- `Symbolic` constructor:

  - The `Symbolic(x::T) where {T<:Number}` constructor now creates a `Term` object from a number `x`, inferring the type of the `Symbolic` object from the type of `x`.
  - The `Symbolic{T}(x::Number) where {T}` constructor now creates a `Term` object with the specified type `T`, converting `x` to that type if necessary.

- convert function:

  - The `convert(::Type{<:Symbolic{T}}, x::Number) where {T}` function now converts a number `x` to a `Term` object of the specified type `T`.

- Tests:

  - The test suite now includes tests for the `Symbolic` constructor and convert function, ensuring correct behavior for various numeric types, including `Int`, `Float64`, `Rational`, and `Complex`.
  - The test suite also includes a test case demonstrating how the convert function works when assigning a number to an array of `Symbolic`.

These changes provide a more convenient and flexible way to create and work with symbolic representations of numbers, and hide the internal `Symbolic` subtype system from external users.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -570,6 +570,57 @@ function basic_similarterm(t, f, args, stype; metadata=nothing)
     end
 end
 
+"""
+$(TYPEDSIGNATURES)
+
+Construct a [`Symbolic`](@ref) object from a number `x`.
+
+This function is used to create a symbolic representation of a number.
+
+```jldoctest
+julia> a = SymbolicUtils.Symbolic(3.14)
+3.14
+
+julia> typeof(a)
+SymbolicUtils.BasicSymbolic{Float64}
+
+julia> b = SymbolicUtils.Symbolic{Real}(6.28)
+6.28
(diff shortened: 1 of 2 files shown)


bowenszhu:
I'm closing this pull request because we've decided to implement the feature using a different approach that will be more efficient and scalable. I'll create a new pull request for this alternative solution soon.
- https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/612

PR #604, opened 2024-06-02
Remove `metadata` kwarg from `maketerm` calls in rewriters.jl

Vaibhavdixit02:
Needed for https://github.com/JuliaSymbolics/Symbolics.jl/pull/1152


Vaibhavdixit02:
This is incorrect

Issue #565, opened 2023-12-07
Setting up coverage in JuliaSymbolics repos.


0x0f0f0f:
Shall we enable codecov? I don't see it around on PRs anymore

shashi:
yes

PR #609, opened 2024-06-08
TermInterface Version 2

0x0f0f0f:


0x0f0f0f:
Have to adjust docs

src/SymbolicUtils.jl
```diff
@@ -16,12 +16,10 @@ using SymbolicIndexingInterface
 import Base: +, -, *, /, //, \, ^, ImmutableDict
 using ConstructionBase
 using TermInterface
-import TermInterface: iscall, isexpr, issym, symtype, head, children,
+import TermInterface: iscall, isexpr, head, children,
                       operation, arguments, metadata, maketerm
 
-const istree = iscall
-Base.@deprecate_binding istree iscall
```
ChrisRackauckas, line 23: where did the deprecation go?

0x0f0f0f: I can restore it, but it looked a bit sketchy... `istree` was defined in TermInterface. It was then renamed to `isexpr`. Why export a function that was deprecated in another package, making it default to another function (and not the one it corresponds to)? Doesn't seem very clean. We could add a deprecation warning in TermInterface if we really want it.

src/types.jl
```diff
@@ -98,6 +98,16 @@ end
...
+"""
+  symtype(x)
+
+Returns the symbolic type of `x`. By default this is just `typeof(x)`.
+Define this for your symbolic types if you want `SymbolicUtils.simplify` to apply rules
+specific to numbers (such as commutativity of multiplication). Or such
+rules that may be implemented in the future.
+"""
+symtype(x) = typeof(x)
 symtype(x::Number) = typeof(x)
```
0x0f0f0f, line 111: We could just leave one method for `symtype`

src/types.jl
```diff
@@ -535,10 +554,12 @@ end
 
 unflatten(t) = t
 
-function TermInterface.maketerm(::Type{<:BasicSymbolic}, head, args, type, metadata)
-    basicsymbolic(head, args, type, metadata)
+function TermInterface.maketerm(T::Type{<:BasicSymbolic}, head, args, metadata)
+    basicsymbolic(head, args, symtype(T), metadata)
 end
 
+symtype(::Type{<:Symbolic{T}}) where T = T
```
0x0f0f0f, line 561: @bowenszhu would like to hear your opinion on this one. Probably not best to have it called `symtype`, but `_symtype_t`?

src/types.jl
ChrisRackauckas: this isn't correct though? It should be the type parameter.

src/SymbolicUtils.jl
ChrisRackauckas: The deprecation is required or this is a breaking update.

src/types.jl
```diff
@@ -535,10 +554,12 @@ end
 
 unflatten(t) = t
 
-function TermInterface.maketerm(::Type{<:BasicSymbolic}, head, args, type, metadata)
-    basicsymbolic(head, args, type, metadata)
+function TermInterface.maketerm(T::Type{<:BasicSymbolic}, head, args, metadata)
+    basicsymbolic(head, args, symtype(T), metadata)
```
0x0f0f0f, line 558: @bowenszhu this is where `_symtype_t` would go

src/types.jl
0x0f0f0f: Yes, it's the type parameter. I can group the definitions together.

src/SymbolicUtils.jl
0x0f0f0f: I think this should be another major. Because I also removed the "compatmaker" rewriter hack below

0x0f0f0f: Which was marked `XXX remove in future major`

docs/src/manual/interface.md
```diff
@@ -13,6 +13,18 @@ You can read the documentation of [TermInterface.jl](https://github.com/JuliaSym
 
 ## SymbolicUtils.jl only methods
 
-`promote_symtype(f, arg_symtypes...)`
+### `symtype(x)`
+
+Returns the symbolic type of `x`. By default this is just `typeof(x)`.
```
bowenszhu, line 18:
```suggestion
Returns the 
[numeric type](https://docs.julialang.org/en/v1/base/numbers/#Standard-Numeric-Types) 
of `x`. By default this is just `typeof(x)`.
```

How about numeric type? Symbolic type is a bit unclear.

bowenszhu, line 18:
```suggestion
Returns the 
[numeric type](https://docs.julialang.org/en/v1/base/numbers/#Standard-Numeric-Types) 
of `x`. By default this is just `typeof(x)`.
```

How about numeric type? Symbolic type is a bit unclear.

docs/src/manual/interface.md
```diff
@@ -13,6 +13,18 @@ You can read the documentation of [TermInterface.jl](https://github.com/JuliaSym
...
+Define this for your symbolic types if you want `SymbolicUtils.simplify` to apply rules
```
bowenszhu, line 19:
```suggestion
Define this for your symbolic types if you want [`SymbolicUtils.simplify`](@ref) to apply rules
```

Add a link to this function.

docs/src/manual/interface.md
```diff
@@ -13,6 +13,18 @@ You can read the documentation of [TermInterface.jl](https://github.com/JuliaSym
...
+specific to numbers (such as commutativity of multiplication). Or such
+rules that may be implemented in the future.
+
+### `issym(x)`
+
+Returns `true` if `x` is a symbol. If true, `nameof` must be defined
+on `x` and must return a Symbol.
```
bowenszhu, line 26:
```suggestion
Returns `true` if `x` is a `Sym`. If `true`, `nameof` must be defined
on `x` and must return a `Symbol`.
```

"symbol" might be misunderstood with Julia [Symbols](https://docs.julialang.org/en/v1/manual/metaprogramming/#Symbols).

src/types.jl
```diff
@@ -98,6 +98,16 @@ end
 ###
 ### Term interface
 ###
+
+"""
+  symtype(x)
```
bowenszhu, line 103:
```suggestion
    symtype(x)
```
A four-space indent is needed so that it is printed as Julia code.
See https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation

src/types.jl
```diff
@@ -98,6 +98,16 @@ end
...
+
+Returns the symbolic type of `x`. By default this is just `typeof(x)`.
+Define this for your symbolic types if you want `SymbolicUtils.simplify` to apply rules
```
bowenszhu, line 106:
```suggestion
    symtype(x)

Returns the numeric type of `x`. By default this is just `typeof(x)`.
Define this for your symbolic types if you want [`SymbolicUtils.simplify`](@ref) to apply rules
```

- A four-space indent is needed so that it is printed as Julia code.
See https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation
- Numeric type might be clearer. https://docs.julialang.org/en/v1/base/numbers/#Standard-Numeric-Types
- Add a cross-reference link to `SymbolicUtils.simplify`. https://documenter.juliadocs.org/stable/man/syntax/#@ref-link

src/types.jl
```diff
@@ -192,7 +202,16 @@ isexpr(s::BasicSymbolic) = !issym(s)
 iscall(s::BasicSymbolic) = isexpr(s)
 
 @inline isa_SymType(T::Val{S}, x) where {S} = x isa BasicSymbolic ? Unityper.isa_type_fun(Val(SymbolicUtils.BasicSymbolic), T, x) : false
+
+"""
+  issym(x)
+
+Returns `true` if `x` is a symbol. If true, `nameof` must be defined
+on `x` and must return a Symbol.
```
bowenszhu, line 210:
```suggestion
    issym(x)

Returns `true` if `x` is a `Sym`. If true, `nameof` must be defined
on `x` and must return a `Symbol`.
```

- A four-space indent is needed so that it is printed as Julia code.
See https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation
- "symbol" might be misunderstood with Julia [Symbols](https://docs.julialang.org/en/v1/manual/metaprogramming/#Symbols).

src/types.jl
bowenszhu: `symtype` is OK.

docs/src/manual/interface.md
0x0f0f0f: Question is, do we really need this default?

src/SymbolicUtils.jl
0x0f0f0f: > > The deprecation hasn't been around long enough to remove. It should be there at least a year or two.
> 
> Makes sense, but that deprecation seems it shouldn't have been there in the first place... I'm happy to reintroduce it but with the correct function.

But I'm not sure whether changing a deprecation is breaking or not? I would remove, but happy to reintroduce if you think this should really be around

docs/src/manual/interface.md
bowenszhu: I think so. While Symbolics.jl primarily focuses on [`BasicSymbolic`](https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/43797eb43b33b8d6b0374eac8cd6f87f94bcf79d/src/types.jl#L26) objects, there are instances where it can be used with other types. 
For example, https://github.com/JuliaSymbolics/Symbolics.jl/issues/1157#issue-2334289624

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
 authors = ["Shashi Gowda"]
-version = "2.0.2"
+version = "2.1.0"
```
0x0f0f0f, line 4: Is this the correct version bump? Or shall this be 3.0.0? Contains breaking

ChrisRackauckas: what is breaking here?

0x0f0f0f: `maketerm` does not take a `symtype` argument anymore because it's inferred from type parameter. `similarterm` was also removed

ChrisRackauckas: oh yes... I don't have the appetite to do another breaking though. If I make the issue will you setup all of the downstream PRs?

0x0f0f0f: Yes. It shouldn't be that hard. I think another breaking is needed then.

ChrisRackauckas: Let's make sure the sorting change is part of the same change then?

src/types.jl
```diff
@@ -535,8 +554,8 @@ end
 
 unflatten(t) = t
 
-function TermInterface.maketerm(::Type{<:BasicSymbolic}, head, args, type, metadata)
-    basicsymbolic(head, args, type, metadata)
+function TermInterface.maketerm(T::Type{<:BasicSymbolic}, head, args, metadata)
+    basicsymbolic(head, args, symtype(T), metadata)
```
YingboMa, line 558: This is not always true. Consider
```julia
ex = x + y
maketerm(typeof(ex), ==, arguments(ex), metadata(ex))
```
The current implementation would derive `x == y` to be `Real` while it should be `Bool`.

YingboMa (changes requested):
Term manipulation could change `symtype`. If we don't want `maketerm` to take `symtype`, we should call `promote_symtype` but that's a dynamic dispatch and expensive. Should we make `symtype` a keyword argument that defaults to `promote_symtype`?

0x0f0f0f:
> Term manipulation could change `symtype`. If we don't want `maketerm` to take `symtype`, we should call `promote_symtype` but that's a dynamic dispatch and expensive. Should we make `symtype` a keyword argument that defaults to `promote_symtype`?

Mmmmh. This is specific to `SymbolicUtils.jl` and `promote_symtype` would not make sense in `TermInterface.jl` because other dependents do not use it.

Project.toml
0x0f0f0f: > Let's make sure the sorting change is part of the same change then?

The one mentioned in tuesday call, right? Defaulting `arguments` to be unsorted and having `sorted_arguments` instead?

0x0f0f0f:
> > Term manipulation could change `symtype`. If we don't want `maketerm` to take `symtype`, we should call `promote_symtype` but that's a dynamic dispatch and expensive. Should we make `symtype` a keyword argument that defaults to `promote_symtype`?
> 
> Mmmmh. This is specific to `SymbolicUtils.jl` and `promote_symtype` would not make sense in `TermInterface.jl` because other dependents do not use it.

@YingboMa the issue you've reported makes sense though and should be fixed.

Since `maketerm` will most likely be a dynamic dispatch anyways, does it make sense to specialize this on the operation (in this case `typeof(==)` to infer the correct symtype?

0x0f0f0f:
> > > Term manipulation could change `symtype`. If we don't want `maketerm` to take `symtype`, we should call `promote_symtype` but that's a dynamic dispatch and expensive. Should we make `symtype` a keyword argument that defaults to `promote_symtype`?
> > 
> > 
> > Mmmmh. This is specific to `SymbolicUtils.jl` and `promote_symtype` would not make sense in `TermInterface.jl` because other dependents do not use it.
> 
> @YingboMa the issue you've reported makes sense though and should be fixed.
> 
> Since `maketerm` will most likely be a dynamic dispatch anyways, does it make sense to specialize this on the operation (in this case `typeof(==)` to infer the correct symtype?

@YingboMa @ChrisRackauckas I'm preparing for conference tomorrow. I can't update at the moment, but happy to do in the next weeks/days if I find some time.

ChrisRackauckas:
@bowenszhu is your https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/615 reliant on this?

bowenszhu:
> @bowenszhu is your #615 reliant on this?

No. It’s not.

0x0f0f0f:
@bowenszhu can we please also add the changes for sorted/unsorted arguments from #615  here? I didn't change many callsites of `arguments`, just low-hanging fruit like `node_count`. This is blocking my work.

0x0f0f0f:
@ChrisRackauckas will break downstream in the same way #615 does. I think it really makes sense to update downstream a single time, otherwise we will need to have 6 PRs (3 for terminterface and 3 for #615), one each for SU, Symbolics and MTK. Please let's synchronise work somehow

ChrisRackauckas:
My understanding is that PR is aiming to not be breaking downstream. Can you rebase on top of it? The deprecation paths and such would be very good to have.

0x0f0f0f:
> The implementation detail of make_term has nothing to do with TermInterface. Just call promote_symtype there to be correct, otherwise, this PR is not acceptable.

@YingboMa will do. Wanted to catch up with #615 first.

0x0f0f0f:
> My understanding is that PR is aiming to not be breaking downstream. Can you rebase on top of it? The deprecation paths and such would be very good to have.

@ChrisRackauckas sure. Will add correct deprecation paths as next step after fixing `maketerm`. At PLDI24 now.

0x0f0f0f:
@YingboMa  I ended up having to do this. I don't really like it, but it would be OK temporarily. Otherwise tests were failing as it was expecting `Number` in some tests and stuff got promoted to `Real` instead.

```
function TermInterface.maketerm(T::Type{<:BasicSymbolic}, head, args, metadata)
    st = symtype(T)
    pst = _promote_symtype(head, args)
    # Use promoted symtype only if not a subtype of the existing symtype of T.
    # This is useful when calling `maketerm(BasicSymbolic{Number}, (==), [true, false])` 
    # Where the result would have a symtype of Bool. 
    # Please see discussion in https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/609 
    # TODO this should be optimized.
    new_st = if pst === Bool 
        pst 
    elseif pst === Any || (st === Number && pst <: st) 
        st
    else 
        pst 
    end 
    basicsymbolic(head, args, new_st, metadata)
end
```

Do you have anything better in mind?

bowenszhu (commented):
Thanks for working on this! I have a couple of suggestions that might make reviewing this PR easier:

1. **Scope:** Could we focus this PR on migrating to TermInterface v1 first? Migrating directly to v2 could be tackled in a separate PR. This would help with reviewing and isolating any potential issues.

2. **Commit https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/609/commits/e9ebd8f56fcc0ddfc7c2510cefd55581f125df9b:** I noticed this commit copies code from PR https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/615 instead of merging it. Would merging be a better approach here to keep the history cleaner and avoid merge conflicts? 

If you're set on migrating directly to v2 and need a hand resolving merge conflicts, let me know! I'm happy to help.

0x0f0f0f:
> Thanks for working on this! I have a couple of suggestions that might make reviewing this PR easier:
> 
>     1. **Scope:** Could we focus this PR on migrating to TermInterface v1 first? Migrating directly to v2 could be tackled in a separate PR. This would help with reviewing and isolating any potential issues.
> 
>     2. **Commit [e9ebd8f](https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/e9ebd8f56fcc0ddfc7c2510cefd55581f125df9b):** I noticed this commit copies code from PR [Optimize `arguments` function by removing sorting #615](https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/615) instead of merging it. Would merging be a better approach here to keep the history cleaner and avoid merge conflicts?
> 
> 
> If you're set on migrating directly to v2 and need a hand resolving merge conflicts, let me know! I'm happy to help.

Hey thanks for the tips. For scope point 1) I guess that it would be incompatible for the way you defined arguments with a sort kwarg. I'll merge everything from master branch

0x0f0f0f:
@ChrisRackauckas @YingboMa @bowenszhu ready for review. I think it'll pass CI

src/SymbolicUtils.jl
0x0f0f0f: Added deprecation

src/polyform.jl
```diff
@@ -176,18 +175,18 @@ isexpr(x::PolyForm) = true
 iscall(x::Type{<:PolyForm}) = true
 iscall(x::PolyForm) = true
 
-function maketerm(::Type{<:PolyForm}, f, args, symtype, metadata)
-    basicsymbolic(t, f, args, symtype, metadata)
+function maketerm(t::Type{<:PolyForm}, f, args, metadata)
+    # TODO: this looks uncovered.
+    basicsymbolic(f, args, nothing, metadata)
 end
```
0x0f0f0f, line 181: This is uncovered by tests and thus had a wrong call to `basicsymbolic`. Can we remove this?

bowenszhu (changes requested):
In the commit https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/609/commits/e9ebd8f56fcc0ddfc7c2510cefd55581f125df9b, you replaced all instances of `arguments` with `sorted_arguments`, which was not fully addressed in the subsequent merge commit https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/609/commits/bfb672d444ce8111569be26d8366654635fbc4a6. The intention behind updating `arguments` was to avoid unnecessary sorting, as it is computationally expensive and not required in many cases.

0x0f0f0f:
> In the commit [e9ebd8f](https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/e9ebd8f56fcc0ddfc7c2510cefd55581f125df9b), you replaced all instances of `arguments` with `sorted_arguments`, which was not fully addressed in the subsequent merge commit [bfb672d](https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/bfb672d444ce8111569be26d8366654635fbc4a6). The intention behind updating `arguments` was to avoid unnecessary sorting, as it is computationally expensive and not required in many cases.

Yeah, I wanted to address them one-by-one, but I guess you did already in your MR? I can align to your changes then.

ChrisRackauckas:
Yes that PR already went one by one to make the choices so just match that

src/code.jl
```diff
@@ -115,7 +115,7 @@ function function_to_expr(op, O, st)
     (get(st.rewrites, :nanmath, false) && op in NaNMathFuns) || return nothing
     name = nameof(op)
     fun = GlobalRef(NaNMath, name)
-    args = map(Base.Fix2(toexpr, st), arguments(O))
+    args = map(Base.Fix2(toexpr, st), sorted_arguments(O))
```
0x0f0f0f, line 118:
```suggestion
    args = map(Base.Fix2(toexpr, st), arguments(O))
```

src/code.jl
```diff
@@ -138,7 +138,7 @@ function function_to_expr(op::Union{typeof(*),typeof(+)}, O, st)
 end
 
 function function_to_expr(::typeof(^), O, st)
-    args = arguments(O)
+    args = sorted_arguments(O)
```
0x0f0f0f, line 141:
```suggestion
    args = arguments(O)
```

src/code.jl
```diff
@@ -151,7 +151,7 @@ function function_to_expr(::typeof(^), O, st)
 end
 
 function function_to_expr(::typeof(SymbolicUtils.ifelse), O, st)
-    args = arguments(O)
+    args = sorted_arguments(O)
```
0x0f0f0f, line 154:
```suggestion
    args = arguments(O)
```

src/code.jl
```diff
@@ -183,7 +183,7 @@ function toexpr(O, st)
         return expr′
     else
         !iscall(O) && return O
-        args = arguments(O)
+        args = sorted_arguments(O)
```
0x0f0f0f, line 186:
```suggestion
        args = arguments(O)
```

src/code.jl
```diff
@@ -693,8 +693,8 @@ end
 function _cse!(mem, expr)
     iscall(expr) || return expr
     op = _cse!(mem, operation(expr))
-    args = map(Base.Fix1(_cse!, mem), arguments(expr))
-    t = similarterm(expr, op, args)
+    args = map(Base.Fix1(_cse!, mem), sorted_arguments(expr))
```
0x0f0f0f, line 696:
```suggestion
    args = map(Base.Fix1(_cse!, mem), arguments(expr))
```

src/code.jl
```diff
@@ -716,7 +716,7 @@ end
 
 function _cse(exprs::AbstractArray)
     letblock = cse(Term{Any}(tuple, vec(exprs)))
-    letblock.pairs, reshape(arguments(letblock.body), size(exprs))
+    letblock.pairs, reshape(sorted_arguments(letblock.body), size(exprs))
```
0x0f0f0f, line 719:
```suggestion
    letblock.pairs, reshape(arguments(letblock.body), size(exprs))
```

src/matchers.jl
```diff
@@ -85,7 +85,7 @@ function matcher(segment::Segment)
 end
 
 function term_matcher(term)
-    matchers = (matcher(operation(term)), map(matcher, arguments(term))...,)
+    matchers = (matcher(operation(term)), map(matcher, sorted_arguments(term))...,)
```
0x0f0f0f, line 88:
```suggestion
    matchers = (matcher(operation(term)), map(matcher, arguments(term))...,)
```

src/ordering.jl
```diff
@@ -84,7 +84,7 @@ function <ₑ(a::BasicSymbolic, b::BasicSymbolic)
     bw = monomial_lt(db, da)
     if fw === bw && !isequal(a, b)
         if _arglen(a) == _arglen(b)
-            return (operation(a), arguments(a)...,) <ₑ (operation(b), arguments(b)...,)
+            return (operation(a), sorted_arguments(a)...,) <ₑ (operation(b), sorted_arguments(b)...,)
```
0x0f0f0f, line 87:
```suggestion
            return (operation(a), arguments(a)...,) <ₑ (operation(b), arguments(b)...,)
```

src/polyform.jl
```diff
@@ -103,7 +103,7 @@ function polyize(x, pvar2sym, sym2term, vtype, pow, Fs, recurse)
         end
 
         op = operation(x)
-        args = arguments(x)
+        args = sorted_arguments(x)
```
0x0f0f0f, line 106:
```suggestion
        args = arguments(x)
```

src/polyform.jl
```diff
@@ -394,7 +391,7 @@ function has_div(x)
 end
 
 flatten_pows(xs) = map(xs) do x
-    ispow(x) ? Iterators.repeated(arguments(x)...) : (x,)
+    ispow(x) ? Iterators.repeated(sorted_arguments(x)...) : (x,)
```
0x0f0f0f, line 394:
```suggestion
    ispow(x) ? Iterators.repeated(arguments(x)...) : (x,)
```

src/rewriters.jl
```diff
@@ -237,15 +221,15 @@ function (p::Walk{ord, C, F, true})(x) where {ord, C, F}
             x = p.rw(x)
         end
         if iscall(x)
-            _args = map(arguments(x)) do arg
+            _args = map(sorted_arguments(x)) do arg
```
0x0f0f0f, line 224:
```suggestion
            _args = map(arguments(x)) do arg
```

src/rewriters.jl
```diff
@@ -237,15 +221,15 @@ function (p::Walk{ord, C, F, true})(x) where {ord, C, F}
...
                 if node_count(arg) > p.thread_cutoff
                     Threads.@spawn p(arg)
                 else
                     p(arg)
                 end
             end
-            args = map((t,a) -> passthrough(t isa Task ? fetch(t) : t, a), _args, arguments(x))
-            t = p.maketerm(x, operation(x), args, metadata=metadata(x))
+            args = map((t,a) -> passthrough(t isa Task ? fetch(t) : t, a), _args, sorted_arguments(x))
```
0x0f0f0f, line 231:
```suggestion
            args = map((t,a) -> passthrough(t isa Task ? fetch(t) : t, a), _args, arguments(x))
```

src/rule.jl
```diff
@@ -121,7 +121,7 @@ getdepth(r::Rule) = r.depth
 
 function rule_depth(rule, d=0, maxdepth=0)
     if iscall(rule)
-        maxdepth = reduce(max, (rule_depth(r, d+1, maxdepth) for r in arguments(rule)), init=1)
+        maxdepth = reduce(max, (rule_depth(r, d+1, maxdepth) for r in sorted_arguments(rule)), init=1)
```
0x0f0f0f, line 124:
```suggestion
        maxdepth = reduce(max, (rule_depth(r, d+1, maxdepth) for r in arguments(rule)), init=1)
```

src/types.jl
```diff
@@ -239,8 +256,8 @@ function _isequal(a, b, E)
     elseif E === POW
         isequal(a.exp, b.exp) && isequal(a.base, b.base)
     elseif E === TERM
-        a1 = arguments(a)
-        a2 = arguments(b)
+        a1 = sorted_arguments(a)
+        a2 = sorted_arguments(b)
```
0x0f0f0f, line 260:
```suggestion
        a1 = arguments(a)
        a2 = arguments(b)
```

src/types.jl
```diff
@@ -669,7 +678,7 @@ const show_simplified = Ref(false)
 isnegative(t::Real) = t < 0
 function isnegative(t)
     if iscall(t) && operation(t) === (*)
-        coeff = first(arguments(t))
+        coeff = first(sorted_arguments(t))
```
0x0f0f0f, line 681:
```suggestion
        coeff = first(arguments(t))
```

src/types.jl
```diff
@@ -701,7 +710,7 @@ end
 function remove_minus(t)
     !iscall(t) && return -t
     @assert operation(t) == (*)
-    args = arguments(t)
+    args = sorted_arguments(t)
```
0x0f0f0f, line 713:
```suggestion
    args = arguments(t)
```

src/utils.jl
```diff
@@ -48,12 +48,12 @@ end
 
 function fold(t)
     if iscall(t)
-        tt = map(fold, arguments(t))
+        tt = map(fold, sorted_arguments(t))
```
0x0f0f0f, line 51:
```suggestion
        tt = map(fold, arguments(t))
```

src/utils.jl
```diff
@@ -74,12 +74,12 @@ _isinteger(x) = (x isa Number && isinteger(x)) || (x isa Symbolic && symtype(x)
 _isreal(x) = (x isa Number && isreal(x)) || (x isa Symbolic && symtype(x) <: Real)
 
 issortedₑ(args) = issorted(args, lt=<ₑ)
-needs_sorting(f) = x -> is_operation(f)(x) && !issortedₑ(arguments(x))
+needs_sorting(f) = x -> is_operation(f)(x) && !issortedₑ(sorted_arguments(x))
```
0x0f0f0f, line 77:
```suggestion
needs_sorting(f) = x -> is_operation(f)(x) && !issortedₑ(arguments(x))
```

src/utils.jl
```diff
@@ -74,12 +74,12 @@ _isinteger(x) = (x isa Number && isinteger(x)) || (x isa Symbolic && symtype(x)
...
 
 # are there nested ⋆ terms?
 function isnotflat(⋆)
     function (x)
-        args = arguments(x)
+        args = sorted_arguments(x)
```
0x0f0f0f, line 82:
```suggestion
        args = arguments(x)
```

src/utils.jl
```diff
@@ -137,29 +137,29 @@ x + 2y
 ```
 """
 function flatten_term(⋆, x)
-    args = arguments(x)
+    args = sorted_arguments(x)
```
0x0f0f0f, line 140:
```suggestion
    args = arguments(x)
```

src/utils.jl
```diff
@@ -137,29 +137,29 @@ x + 2y
...
     # flatten nested ⋆
     flattened_args = []
     for t in args
         if iscall(t) && operation(t) === (⋆)
-            append!(flattened_args, arguments(t))
+            append!(flattened_args, sorted_arguments(t))
```
0x0f0f0f, line 145:
```suggestion
            append!(flattened_args, arguments(t))
```

src/utils.jl
```diff
@@ -137,29 +137,29 @@ x + 2y
...
             push!(flattened_args, t)
         end
     end
-    maketerm(typeof(x), ⋆, flattened_args, symtype(x), metadata(x))
+    maketerm(typeof(x), ⋆, flattened_args, metadata(x))
 end
 
 function sort_args(f, t)
-    args = arguments(t)
+    args = sorted_arguments(t)
```
0x0f0f0f, line 154:
```suggestion
    args = arguments(t)
```

src/utils.jl
```diff
@@ -182,12 +182,12 @@ Base.length(l::LL) = length(l.v)-l.i+1
 Base.length(t::Term) = length(arguments(t)) + 1 # PIRACY
 Base.isempty(t::Term) = false
 @inline car(t::Term) = operation(t)
-@inline cdr(t::Term) = arguments(t)
+@inline cdr(t::Term) = sorted_arguments(t)
```
0x0f0f0f, line 185:
```suggestion
@inline cdr(t::Term) = arguments(t)
```

src/utils.jl
```diff
@@ -182,12 +182,12 @@ Base.length(l::LL) = length(l.v)-l.i+1
...
 
 @inline car(v) = iscall(v) ? operation(v) : first(v)
 @inline function cdr(v)
     if iscall(v)
-        arguments(v)
+        sorted_arguments(v)
```
0x0f0f0f, line 190:
```suggestion
        arguments(v)
```

src/utils.jl
```diff
@@ -200,7 +200,7 @@ end
     if n === 0
         return ll
     else
-        iscall(ll) ? drop_n(arguments(ll), n-1) : drop_n(cdr(ll), n-1)
+        iscall(ll) ? drop_n(sorted_arguments(ll), n-1) : drop_n(cdr(ll), n-1)
```
0x0f0f0f, line 203:
```suggestion
        iscall(ll) ? drop_n(arguments(ll), n-1) : drop_n(cdr(ll), n-1)
```

0x0f0f0f:
> Yes that PR already went one by one to make the choices so just match that

I went to the PRs side by side and reverted the non-fundamental `sorted_arguments`. However I think that many of these cases will cause correctness issues in the futures, and may be already holding flakily because the possible permutations of small dictionaries (multisets with unordered dicts) are few. The fuzzing is supposed to cover this but I'm not sure how many functions the actual tests check

src/types.jl
```diff
@@ -196,7 +204,16 @@ isexpr(s::BasicSymbolic) = !issym(s)
...
+
+"""
+    issym(x)
+
+Returns `true` if `x` is a `Sym`. If true, `nameof` must be defined
+on `x` and must return a `Symbol`.
+"""
+issym(x) = false
 issym(x::BasicSymbolic) = isa_SymType(Val(:Sym), x)
+
```
bowenszhu, line 216:
```suggestion
issym(x) = isa_SymType(Val(:Sym), x)
```

Type checking is done in `isa_SymType`. 
https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/d009ca8e986b12982dce13794218ea6b632da617/src/types.jl#L198

src/polyform.jl
bowenszhu: I noticed this pull request addresses a few different things and has been open for a while now. To help get it merged more quickly, I think it would streamline the review process if we focused on one issue per PR.

You can always open new issues or pull requests for the other changes later. This will make it easier for us to:
- Review the changes effectively: Focusing on a single issue per PR allows for more thorough reviews and reduces the chance of overlooking potential issues.
- Merge changes faster: Smaller, focused PRs are generally quicker to review and merge.
- Maintain a cleaner commit history: Keeping commits focused makes it easier to understand the evolution of the codebase.

0x0f0f0f:
Ready for review again

0x0f0f0f:
@ChrisRackauckas @bowenszhu any news for this MR?

src/polyform.jl
0x0f0f0f: @bowenszhu unfortunately this is the minimal changeset required to bump TI to 2.0 . i like smaller PRs as well

0x0f0f0f:
> In the commit [e9ebd8f](https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/e9ebd8f56fcc0ddfc7c2510cefd55581f125df9b), you replaced all instances of `arguments` with `sorted_arguments`, which was not fully addressed in the subsequent merge commit [bfb672d](https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/bfb672d444ce8111569be26d8366654635fbc4a6). The intention behind updating `arguments` was to avoid unnecessary sorting, as it is computationally expensive and not required in many cases.

@bowenszhu this was addressed

0x0f0f0f:
Hey guys @ChrisRackauckas @bowenszhu ping

src/types.jl
```diff
@@ -539,8 +555,22 @@ end
...
+    # Where the result would have a symtype of Bool. 
+    # Please see discussion in https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/609 
+    # TODO this should be optimized.
+    new_st = if pst === Bool 
+        pst 
+    elseif pst === Any || (st === Number && pst <: st) 
+        st
+    else 
+        pst 
+    end
```
bowenszhu, line 572: This looks good! To ensure we're fully testing the new cases, could we add tests to reach 100% branch coverage here?

bowenszhu (changes requested):

src/types.jl
0x0f0f0f: Do you have any idea for a test structure? Also #565

bowenszhu: Since codecov is currently disabled, we probably have to do a thorough manual review of the code and tests to ensure all branches are adequately covered.
IMG_7475

0x0f0f0f:
@bowenszhu added tests covering new maketerm symtype propagation in latest commit

bowenszhu (approved):

Issue #612, opened 2024-06-12
Add `Const(::Any)` to make `BasicSymbolic` fully closed


YingboMa:
#611 is the prerequisite so that we can use recursive types. Instead of `Vector{Any}` or `Dict{Any, X}` we will have `Vector{BasicSymbolic}` and `Dict{BasicSymbolic, X}`.

Issue #611, opened 2024-06-12
Migrate to Expronicon.jl v0.8 from Unityper.jl


YingboMa:
Unityper.jl doesn't support recursive type and it also doesn't support powerful pattern matching with MLStyle.jl.

Note that we have to use Expronicon.jl v0.8 and not any other release because later version cause massive invalidations.

CC: @Roger-luo

Roger-luo:
Well I do realize it's a bad idea overloading getproperty to provide namespace syntax. So I'm still working on a refactor of Expronicon sum type. I have a working version inside Liang.jl but still WIP splitting that out and stabilizing.

bowenszhu:
I'm running into an issue with using parametric types like `BasicSymbolic{T}` with Expronicon v0.10.8. It seems that Expronicon doesn't yet support them, as demonstrated by the following error:
```
module SymbolicADT
using Expronicon.ADT: @adt

@adt BasicSymbolic{T} begin end

end # module
```
Error:
```
ERROR: LoadError: ArgumentError: generic ADT is not supported yet
...
```
Is there a workaround for this, or is support for parametric types planned for a future release?

Thanks!

Roger-luo:
Yes it doesn't support such for the same reason as Unityper - if we obey C tagged union convention this will require Julia compiler to support generated struct (as pointed out in a previous discussion under JuliaLang/julia)

However, Mason have demonstrated that without this convention we can achieve better performance by just using julia's Union type in SumTypes.jl. I'm planning to adopt this solution and rewrite MLStyle on top of this. The new package is named Moshi.jl stands for the name pattern in Chinese. 

But as you may know I just graduated recently so I will be in vacation and moving for the summer. So I'm not sure how much time I have to push this forward this year.

bowenszhu:
@YingboMa 
Do you have any recommendations on how to handle parametric types with Expronicon for the time being?

YingboMa:
Since `T` in `Symbolic{T}` is only a tag for for symtype. We could just do
```julia
struct BasicSymbolic{T}
    x::BasicSymbolicImpl
end
```
and use `Expronicon` to implement `BasicSymbolicImpl`.

bowenszhu:
@YingboMa 
Thanks for the suggestion! That's a smart workaround for the parametric type limitation.

Roger-luo:
[...]

PR #617, opened 2024-06-18
Migrate to Expronicon.jl, add `Const(::Any)`

bowenszhu:
- fix https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/611
- fix https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/612
- fix https://github.com/JuliaSymbolics/Symbolics.jl/issues/853
- fix https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/601


src/types.jl
```diff
@@ -1,85 +1,74 @@
...
+    struct Term
+        f::Any
+        arguments::Vector{BasicSymbolic}
     end
-    struct Term{T} <: BasicSymbolic{T}
-        f::Any                 = identity  # base/num if Pow; issorted if Add/Dict
-        arguments::Vector{Any} = EMPTY_ARGS
-        hash::RefValue{UInt}   = EMPTY_HASH
+    struct Add
+        coeff::BasicSymbolic
```
bowenszhu, line 18: @YingboMa Should we restrict the type of the `coeff` field in `Add` and `Mul` to `Const` here?

bowenszhu:
I'm posting an update on this PR. While the migration to Expronicon offers potential benefits like recursive algebraic data types and wrapping numerical values in BasicSymbolic, the work involved has proven more engineering intensive than initially anticipated. It has resulted in numerous broken tests, both within SymbolicUtils and in downstream packages, and debugging these requires a significant time investment per test.

At this point, I don't have a clear timeline for completing this migration, and the performance implications are still uncertain. Due to other priorities, I'm currently putting this work on hold.

I'm happy to hand this off to anyone interested in pursuing it. The PR contains the current state of the work, and I'm available to answer questions and provide context. Please comment here if you'd like to take over or discuss further. Any help would be greatly appreciated!

akirakyle:
Hi @bowenszhu! I originally wasn't intending to spend any time on this, I just came across it out of my questions in https://github.com/JuliaSymbolics/Symbolics.jl/issues/1354 along with my interest in implementing non-commutative symbolics for applications quantum info. However, I had a some free time this past weekend to take at look at this and also all the various approaches to implementing ADT in julia. So I thought a good exercise for myself to better understand how the SymbolicUtils works would be to try to do the same thing this PR does, except with Moshi.jl. The result of my hacking on this this last weekend is #676.

Anyways, I don't necessarily have the bandwidth to move forward on this very quickly. But if it does seem like something that may turn out to benefit julia symbolics and get merged I do have interest in continuing to pursue this.

PR #623, opened 2024-07-17
Fix faulty istree deprecation

termi-official:
This should essentially fix https://github.com/SciML/ModelingToolkit.jl/issues/2827 . The issue was that istree has been marked as deprecated, but istree is also the same function as iscall, causing Aqua to fail.


termi-official:
`IntegrationTest / Symbolics.jl/All (pull_request)` also fails on master.

PR #624, opened 2024-07-17
Fix metadata prop

YingboMa:
`-(-x)` will simplify back to `x` and setting metadata on the simplified result is illegal.

Fixes https://github.com/SciML/ModelingToolkit.jl/issues/2855

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -553,14 +553,38 @@ function basicsymbolic(f, args, stype, metadata)
         T = _promote_symtype(f, args)
     end
     if T <: LiteralReal
-        Term{T}(f, args, metadata=metadata)
-    elseif T <: Number && (f in (+, *) || (f in (/, ^) && length(args) == 2)) && all(x->symtype(x) <: Number, args)
-        res = f(args...)
-        if res isa Symbolic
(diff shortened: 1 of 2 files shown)


YingboMa:
Tests failures seem unrelated.

ChrisRackauckas:
Is https://github.com/SciML/ModelingToolkit.jl/issues/2855 related?

YingboMa:
It should if fix it as well.
```julia
julia> buggy(-1) # [-B₊x(t) ~ 0], as expected
1-element Vector{Equation}:
 -B₊x(t) ~ 0

julia> buggy(0)  # [0 ~ 0], as expected
1-element Vector{Equation}:
 0 ~ 0

julia> buggy(1)  # [x(t) ~ 0], but expected [B₊x(t) ~ 0]
1-element Vector{Equation}:
 B₊x(t) ~ 0

julia> buggy(2)  # [2B₊x(t) ~ 0], as expected
1-element Vector{Equation}:
 2B₊x(t) ~ 0
```

ChrisRackauckas:
Well that explains that. Perfect.

PR #627, opened 2024-08-02
fix typo of docs `rule`

oameye:


bowenszhu (approved):
Thanks for catching and fixing this typo!

PR #634, opened 2024-08-19
fix: avoid `ConstructionBase.setproperties` ambiguity

AayushSabharwal:
Fixes [this](https://github.com/SciML/ModelingToolkit.jl/actions/runs/10447617309/job/28926807757?pr=2945#step:6:627) ambiguity. The [`ConstructionBase` docs](https://juliaobjects.github.io/ConstructionBase.jl/stable/#ConstructionBase.setproperties) explicitly recommend adding a method to `setproperties` (`setproperties_object` is undocumented internals) and type-annotating `patch::NamedTuple`.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -89,7 +89,7 @@ const SIMPLIFIED = 0x01 << 0
 #@inline is_of_type(x::BasicSymbolic, type::UInt8) = (x.bitflags & type) != 0x00
 #@inline issimplified(x::BasicSymbolic) = is_of_type(x, SIMPLIFIED)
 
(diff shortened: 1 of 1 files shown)


devmotion:
This change also broke old SymbolicUtils versions such as eg. 1.5.1 or 1.7.0 (which I have to keep using for now due to compatibility with MTK v8: https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/595). Can we backport this fix to older versions?

ChrisRackauckas:
Yes we can backport. Just to v2?

devmotion:
Backporting to v1 is necessary for MTK 8.

ChrisRackauckas:
Backported all the way to v0.19

devmotion:
Thank you!

PR #635, opened 2024-08-21
adjoint(y::Symbolic{<:Number})

ChristophHotter:
I added the method  `adjoint(y::Symbolic{<:Number}) = conj(s)`. 
This is what Julia does for example with `adjoint(1im)`.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -160,36 +160,13 @@ for f in [identity, one, zero, *, +, -]
 end
 
 promote_symtype(::typeof(Base.real), T::Type{<:Number}) = Real
+Base.real(s::Symbolic{<:Number}) = islike(s, Real) ? s : term(real, s)
 promote_symtype(::typeof(Base.conj), T::Type{<:Number}) = T
+Base.conj(s::Symbolic{<:Number}) = islike(s, Real) ? s : term(conj, s)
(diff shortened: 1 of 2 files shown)


bowenszhu (changes requested):
Thanks for the PR! Could you add some test?

ChristophHotter:
Yes, I added tests for Real and Complex syms.

bowenszhu (changes requested):
[`Base.adjoint`](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Base.adjoint) implements lazy evaluation.
https://github.com/JuliaLang/julia/blob/6f3fdf7b36250fb95f512a2b927ad2518c07d2b5/stdlib/LinearAlgebra/src/adjtrans.jl#L280

```julia
A = [3+2im 9+2im; 0  0]
adjoint(adjoint(A)) === A
```
`adjoint` of an `adjoint` is the original variable. 
We should mimic the same behavior.

bowenszhu (approved):
Looks good. Thank you!

https://github.com/JuliaLang/julia/blob/6f3fdf7b36250fb95f512a2b927ad2518c07d2b5/base/number.jl#L225

ChristophHotter:
Thanks!

PR #641, opened 2024-08-28
lazy evaluation of rational powers

mxhbl:
Rational powers are currently eagerly evaluated, so that, for example, 
```
using SymbolicUtils
@syms x
(2x) ^ (1//2) == 1.414… * x^(1//2)
simplify(term(^, 2, 1//2)) == 1.414…
```
and so on. While this might be intentional, it often causes precision loss and makes many analytical expressions rather ugly.

This PR adds branches to `basicsymbolic` and `^` to prevent evaluation of powers if the base is a literal number and the exponent is a rational. Integer or float exponents still get evaluated.
With this PR, we therefore have behavior like
```
(2x) ^ (1//2) ==  2^(1//2) * x^(1//2)
simplify(term(^, 2, 1//2)) == 2^(1//2)
simplify(term(^, 1//3, 2)) == 1//9
simplify(term(^, 2, 0.5)) == 1.414…
```

However, there is a small quirk. Currently `(2x) ^ 2` results in `(4//1) * x^2`, i.e. the integer prefactor is converted to a rational, due to the behavior of the `unstable_pow` function. I am not sure if this is intentional and I did not follow this conversion in the code in this PR. This means with this PR and a rational exponent, we have `(2x) ^ (2//1) == 4 * x^(2//1)`. I guess if this PR gets merged, this behavior should be made consistent, one way or another.

I am curious if this all sounds like a reasonable change, or if there other reasons why immediate evaluation of all powers is preferable — please let me know what you think!

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -608,9 +608,17 @@ function basicsymbolic(f, args, stype, metadata)
             end
             res
         elseif f == (^) && length(args) == 2
-            res = args[1] ^ args[2]
-            if ispow(res)
-                @set! res.metadata = metadata
+            if args[2] isa Rational && !(args[1] isa Symbolic)
+                if !isinteger(args[2])
+                    @goto FALLBACK
+                end
+                integer_type = only(typeof(args[2]).parameters)
+                res = args[1] ^ convert(integer_type, args[2])
+            else
+                res = args[1] ^ args[2]
+                if ispow(res)
+                    @set! res.metadata = metadata
+                end
             end
             res
         else
@@ -1210,9 +1218,21 @@ function ^(a::SN, b)
     elseif b isa Number && b < 0
         Div(1, a ^ (-b))
     elseif ismul(a) && b isa Number
-        coeff = unstable_pow(a.coeff, b)
+        new_dict = mapvalues((k, v) -> b*v, a.dict)
+        if b isa Rational
+            if isinteger(b)
+                integer_type = only(typeof(b).parameters)
+                coeff = a.coeff ^ convert(integer_type, b)
+            else
+                coeff = 1
+                merge!(new_dict, Dict(term(^, a.coeff, b) => 1))
+            end
+        else
+            coeff = unstable_pow(a.coeff, b)
+        end
+
         Mul(promote_symtype(^, symtype(a), symtype(b)),
-            coeff, mapvalues((k, v) -> b*v, a.dict))
+            coeff, new_dict)
     else
         Pow(a, b)
     end
--- a/test/rulesets.jl
+++ b/test/rulesets.jl
@@ -52,6 +52,14 @@ end
     @test simplify(Term(zero, [a])) == 0
     @test simplify(Term(zero, [b + 1])) == 0
     @test simplify(Term(zero, [x + 2])) == 0
+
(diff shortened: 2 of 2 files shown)


bowenszhu (changes requested):
Thanks for the PR! Could you add some tests for `^(a::SN, b)` only without `simplify`?

mxhbl:
Done! Let me know if there is anything else.

src/types.jl
```diff
@@ -608,9 +608,17 @@ function basicsymbolic(f, args, stype, metadata)
...
+                    @goto FALLBACK
+                end
+                integer_type = only(typeof(args[2]).parameters)
+                res = args[1] ^ convert(integer_type, args[2])
+            else
+                res = args[1] ^ args[2]
+                if ispow(res)
+                    @set! res.metadata = metadata
+                end
             end
```
bowenszhu, line 622: `metadata` should be set in both cases.

```suggestion
            end
            if ispow(res)
                @set! res.metadata = metadata
            end
```

bowenszhu (changes requested):

src/types.jl
```diff
@@ -608,9 +608,17 @@ function basicsymbolic(f, args, stype, metadata)
...
         elseif f == (^) && length(args) == 2
-            res = args[1] ^ args[2]
-            if ispow(res)
-                @set! res.metadata = metadata
+            if args[2] isa Rational && !(args[1] isa Symbolic)
+                if !isinteger(args[2])
+                    @goto FALLBACK
+                end
+                integer_type = only(typeof(args[2]).parameters)
+                res = args[1] ^ convert(integer_type, args[2])
```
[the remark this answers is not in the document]
mxhbl: Oh yeah that's much easier.

src/types.jl
```diff
@@ -1210,9 +1218,21 @@ function ^(a::SN, b)
     elseif b isa Number && b < 0
         Div(1, a ^ (-b))
     elseif ismul(a) && b isa Number
-        coeff = unstable_pow(a.coeff, b)
+        new_dict = mapvalues((k, v) -> b*v, a.dict)
+        if b isa Rational
+            if isinteger(b)
+                integer_type = only(typeof(b).parameters)
+                coeff = a.coeff ^ convert(integer_type, b)
```
bowenszhu, line 1225:
```suggestion
                coeff = a.coeff ^ numerator(b)
```

bowenszhu (changes requested):

src/types.jl
```diff
@@ -1210,9 +1218,20 @@ function ^(a::SN, b)
...
         Div(1, a ^ (-b))
     elseif ismul(a) && b isa Number
-        coeff = unstable_pow(a.coeff, b)
+        new_dict = mapvalues((k, v) -> b*v, a.dict)
+        if b isa Rational
+            if isinteger(b)
+                coeff = a.coeff ^ numerator(b)
+            else
+                coeff = 1
+                merge!(new_dict, Dict(term(^, a.coeff, b) => 1))
```
bowenszhu, line 1227: Constructing a `Dict` is unnecessarily expensive here, as it allocates vectors of length 16.
https://github.com/JuliaLang/julia/blob/6f3fdf7b36250fb95f512a2b927ad2518c07d2b5/base/dict.jl#L68-L71

PR #642, opened 2024-08-29
Nested associative-commutative pattern matching

zengmao:
This PR enables full (nested) associative-commutative pattern matching for all sub-expressions with `+` or `*` operations, when the user passes an option `fullac` to the `@rule` macro. This solves the problem that `@acrule` only applies associativity-commutativity at the top level.

The algorithm is currently simplistic, meant to be a "minimum viable product" rather than an optimized implementation. The pattern matcher tries up to $n!$ permutations of arguments for any sub-expression with the operation `+` or `*`, when attempting to find matches. The core algorithm is implemented in `matcher.jl` in the function `function term_matcher(term, fullac_flag = false)`.

Here are examples which show capabilities beyond those of `@acrule`:
```
(@rule ~a + ~a*~b => ~a * (1+~b) fullac)(b + a*b) # result: b * (1+a)
(@rule ~a*~b + ~a*~c => ~a * (~b+~c) fullac)(a*b + b*c) # result: b * (a+c)
```

### DIFF
--- a/src/matchers.jl
+++ b/src/matchers.jl
@@ -6,13 +6,18 @@
 # 3. Callback: takes arguments Dictionary × Number of elements matched
 #
 function matcher(val::Any)
-    iscall(val) && return term_matcher(val)
+    matcher(val, false)
+end
+
+# `fullac_flag == true` enables fully nested associative-commutative pattern matching
+function matcher(val::Any, fullac_flag)
+    iscall(val) && return term_matcher(val, fullac_flag)
     function literal_matcher(next, data, bindings)
         islist(data) && isequal(car(data), val) ? next(bindings, 1) : nothing
     end
 end
 
-function matcher(slot::Slot)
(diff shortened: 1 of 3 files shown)


zengmao:
It would be useful to hear some feedback about:

- Since Metatheory.jl potentially can handle associativity-commutativity fully in the alternative e-graph approach, does SymbolicsUtil still need a classic term rewriter that's aware of associativity-commutativity beyond the top level? Or do you prefer to keep the classical rewriter in its current state, rather than adding features like this PR?
- Is the user interface, i.e. an extra option for `@rule`, acceptable? Or would it be better to make a new macro like `@fullac_rule`?
-  Is there a risk of a resource-intensive feature getting used too heavily by users and causing performance degradation? Maybe the users just need to be warned to prefer ordinary `@rule` and `@acrule` when possible.

PR #657, opened 2024-10-13
fix: `similarterm` is exported but not defined

avik-pal:


bowenszhu (approved):

avik-pal:
Can we do a patch release for this?

bowenszhu:
Sure. OK.

Issue #614, opened 2024-06-12
Hash consing


YingboMa:
We can add a global thread-safe dictionary to de-duplicate terms in memory. This will guarantee the memory cost is the same as an efficient DAG representation while preserving the tree data structure.

0x0f0f0f:
It's basically an e-graph and we can already use Metatheory.jl for that.

PR #658, opened 2024-10-14
Implement hash consing for `Sym`

bowenszhu:
- #614

This PR implements hash consing for `Sym` as the first step towards applying hash consing across SymbolicUtils.jl.

### Motivation:

Hash consing enables efficient memory usage and potential performance improvements by ensuring that only one unique copy of a symbolic expression exists in memory.

### Implementation Details:

- A [`WeakValueDict`](https://github.com/twavv/WeakValueDicts.jl) is used to store existing symbolic objects. This choice ensures that symbolic objects without strong references are eligible for garbage collection.
- The `BasicSymbolic` struct is modified to be `mutable` to comply with the requirements of `WeakValueDict` (due to how Julia implements finalizers).
- A new "hash" function, `hash2`, is introduced to incorporate `symtype` into the hash calculation. This approach was chosen over modifying the existing `hash` function to avoid breaking many tests because ordering is dependent on hash values.
- A new `isequal2` function is added for metadata comparison in addition to `Base.isequal`, because directly modifying `Base.isequal` breaks numerous tests in ModelingToolkit.jl.

Future PRs will apply hash consing to other `BasicSymbolic` subtypes.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -309,6 +309,11 @@ function Base.hash(s::BasicSymbolic, salt::UInt)::UInt
     end
 end
 
+hash2(s::BasicSymbolic) = hash2(s, zero(UInt))
+function hash2(s::BasicSymbolic{T}, salt::UInt)::UInt where {T}
+    hash(T, hash(s, salt))
+end
+
 ###
 ### Constructors
 ###
--- a/src/types.jl
+++ b/src/types.jl
@@ -332,6 +332,18 @@ function Base.hash(s::BasicSymbolic, salt::UInt)::UInt
     end
 end
 
+"""
+$(TYPEDSIGNATURES)
+
+Calculates a hash value for a `BasicSymbolic` object, incorporating both its metadata and 
+symtype.
+
+This function provides an alternative hashing strategy to `Base.hash` for `BasicSymbolic` 
+objects. Unlike `Base.hash`, which only considers the expression structure, `hash2` also 
+includes the metadata and symtype in the hash calculation. This can be beneficial for hash 
+consing, allowing for more effective deduplication of symbolically equivalent expressions 
+with different metadata or symtypes.
+"""
 hash2(s::BasicSymbolic) = hash2(s, zero(UInt))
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Looks reasonable enough and the implementation is clean. Are there performance benchmarks to see if it's going on the right direction?

AayushSabharwal:
Considering this matches hashes to symbolic variables, wouldn't it also need to handle hash collisions? If two `Sym`s end up with the same hash, it would return a different one than what the user expects.

Also, if hashconsing is using a custom hash function could it also hash the metadata and not require that it be empty?

bowenszhu:
requesting review @AayushSabharwal @ChrisRackauckas

src/types.jl
```diff
@@ -77,6 +77,8 @@ function exprtype(x::BasicSymbolic)
     end
 end
 
+wvd = WeakValueDict{UInt, BasicSymbolic}()
```
AayushSabharwal, line 80: Is there a benefit to not making this `const`? Wouldn't it lead to non-const global issues?

bowenszhu: Updated. Thank you!

src/types.jl
```diff
@@ -265,6 +271,26 @@ function _isequal(a, b, E)
...
+and ignores metadata. This does not help deal with hash collisions when metadata is 
+relevant for distinguishing expressions, particularly in hashing contexts. This function
+provides a stricter equality check that includes metadata comparison, preventing
+such collisions.
+
+Modifying `Base.isequal` directly breaks numerous tests in `SymbolicUtils.jl` and
+downstream packages like `ModelingToolkit.jl`, hence the need for this separate
+function.
+"""
+function isequal2(a::BasicSymbolic, b::BasicSymbolic)::Bool
```
ChrisRackauckas, line 290: needs a better name, `complete_isequal`, `isequal_with_metadata`, or something .

bowenszhu: Changed to `isequal_with_metadata`.

Issue #665, opened 2024-10-25
`isequal` for `BasicSymbolic` should consider metadata


bowenszhu:
The current implementation of `Base.isequal` for `BasicSymbolic` objects doesn't take into account potential metadata differences. This can lead to situations where two symbolically equivalent expressions with different metadata are considered equal, which might not be the desired behavior.

https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/f9b0ade4a407b6c090ea8ce39fc9ca4ef238826e/src/types.jl#L241-L266

PR #669, opened 2024-11-04
Enhancement: Include metadata in `BasicSymbolic` `isequal` comparison

bowenszhu:
- Fix https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/665

This pull request enhances the `isequal` function for `BasicSymbolic` to incorporate metadata in the comparison.  Previously, metadata was ignored when checking for equality between two symbolic objects. This could lead to incorrect results when mathematically equivalent sysmbolic objects have different metadata, especially for #658.

This change ensures that two `BasicSymbolic` objects are considered equal only if they have the same expression type, the same symtypes, and identical metadata. This PR specifically supports the work done in #658 for internally handling hash collision.


bowenszhu:
Following up on @AayushSabharwal's point from the Oct 31 meeting: including metadata in `BasicSymbolic` `isequal` is breaking ModelingToolkit integration tests. Should we prioritize metadata compatibility or maintaining the current integration tests? Need a decision here to move forward.

AayushSabharwal:
Adding metadata equality to `isequal` would be breaking for SU and require a massive change to MTK to support. In the October 31st meeting, I was suggesting considering metadata specifically for hashconsing. Similar to the `hash2` function, an `isequal2` could be used to check for hash collisions. Effectively, something like:

```julia
h = hash2(sym)
cached = get(wvd, h, nothing)
if cached === nothing
    wvd[h] = sym
elseif isequal2(cached, sym)
    return cached
else
    return sym
end
```

PR #673, opened 2024-11-11
Apply hash consing to all `BasicSymbolic` subtypes

Blablablanca:
- resolve https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/614

- following https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/658

(UROP under @bowenszhu)


bowenszhu:
The last commit https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/673/commits/f2d779194cd3d4863f366455e82ecebee358446e implements the change to `hash2` to ensure that `a [with metadata] + b` and `a [no metadata] + b` have different hashes. The recursive approach now incorporates metadata and symtype information, correctly distinguishing these cases. The `hash2` function could be optimized further in a future PR, for example, by adding a field to the `BasicSymbolic` struct to cache hash values and avoid redundant computations. The ordering of symbolic objects is not dependent on `hash2`, allowing for future refactoring of this function.

ChrisRackauckas (approved):

PR #680, opened 2024-12-12
Add rules to distribute conj over sum and prod

benneti:
distributing conj makes it easier to apply the remaining rules


benneti:
As this is my first PR in this repo pointers are very welcome. 
To slighly expand on my thinking here: at the moment complex conjugating a longer expression including real variables stops the simplification. In a lot of cases this simplification would avoid this.
A very simple example `@syms x::Real y::Number; conj(x*y)*x*y |> simplify
` which could not be simplified before an now becomes `(x^2)*y*conj(y)`.

For me locally rulesets tests are all green.

PR #686, opened 2025-01-08
Specialize NaNMath.pow to allow for ::Int ^ ::Int -> ::Int

ChrisRackauckas:
NaNMath does not handle integer dispatches. This fixes it so that a wrapper function handles the integer dispatch, restoring the behavior of ^

### DIFF
--- a/src/code.jl
+++ b/src/code.jl
@@ -9,7 +9,7 @@ export toexpr, Assignment, (←), Let, Func, DestructuredArgs, LiteralExpr,
 import ..SymbolicUtils
 import ..SymbolicUtils.Rewriters
 import SymbolicUtils: @matchable, BasicSymbolic, Sym, Term, iscall, operation, arguments, issym,
-                      symtype, sorted_arguments, metadata, isterm, term, maketerm
+                      symtype, sorted_arguments, metadata, isterm, term, maketerm, pow
 import SymbolicIndexingInterface: symbolic_type, NotSymbolic
 
 ##== state management ==##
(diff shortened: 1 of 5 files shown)


src/code.jl
```diff
@@ -146,12 +146,12 @@ function function_to_expr(op::typeof(^), O, st)
             return toexpr(Term(inv, Any[ex]), st)
         else
             args = Any[Term(inv, Any[ex]), -args[2]]
-            op = get(st.rewrites, :nanmath, false) ? op : NaNMath.pow
+            op = get(st.rewrites, :nanmath, false) ? op : pow
```
devmotion, line 149: This seems wrong (was introduced in #671), shouldn't it be
```suggestion
            op = get(st.rewrites, :nanmath, false) === true ? pow : op
```

devmotion (commented):
I just noticed the `Int` problem as well. Wouldn't it be easier to just never rewrite `^` to `NaNMath.pow` if `args[2] isa Int`? There shouldn't be any numerical issues anyway in this case.

ChrisRackauckas:
In theory everyone's symbolic types should match the types they end up passing to the function. In practice I think it's more robust to not assume that, and since it's practically zero overhead to not assume that (since this dispatch should compile out in almost all cases of symbolic generated functions), I would take the more robust route.

devmotion:
On a second thought, I guess we should just add `pow(x::Real, y::Int) = x^y` to NaNMath?

devmotion:
I opened https://github.com/JuliaMath/NaNMath.jl/pull/83.

ChrisRackauckas:
Yeah that's a more general solution.

PR #687, opened 2025-01-08
Do not rewrite `^(_, ::Int)` with `NaNMath.pow`

### LINKED ISSUE #685: using NaNmath.pow changes the type of integer symbolics
`NaNmath.pow(::Integer, ::Integer)` generates floating point numbers, thereby breaking anything Symbolic that previously relied on integer symbolic powers remaining integers (for example, symbolic stoichiometry is now broken in Catalyst with SymbolicUtils 2.8).

devmotion:
Fixes #685 by not rewriting `^(_, ::Int)` which is safe even for negative numbers.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
 authors = ["Shashi Gowda"]
-version = "3.8.1"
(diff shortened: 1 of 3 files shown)


devmotion:
I just saw https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/686#issuecomment-2578856336, so it seems #686 would be safer.

PR #688, opened 2025-01-09
Faster common subexpression elimination

bowenszhu:


AayushSabharwal:
Time-dependent symbolics are treated as function calls. This doesn't affect codegen, but is something I think is worth acknowledging:

```julia
julia> @syms t x(t) y(t)
julia> x = x(t)
julia> y = y(t)

julia> topological_sort((x + y) * sin(x + y))
5-element Vector{SymbolicUtils.Code.Assignment}:
 SymbolicUtils.Code.Assignment(var"##cse#230", x(t))
 SymbolicUtils.Code.Assignment(var"##cse#231", y(t))
 SymbolicUtils.Code.Assignment(var"##cse#232", var"##cse#230" + var"##cse#231")
 SymbolicUtils.Code.Assignment(var"##cse#233", sin(var"##cse#232"))
 SymbolicUtils.Code.Assignment(var"##cse#234", var"##cse#232"*var"##cse#233")
```

Note how it creates assignments for `x(t)` and `y(t)`. This does add to the size of the generated code, but it only scales linearly with the number of variables so that shouldn't be a problem.

The nice part of this is it handles CSE for DDEs auto-magically:

```julia
julia> @syms dde(t)

julia> topological_sort(dde(t + t^2 + 1) * t^2 * sin(dde(t + t^2 + 1)))
5-element Vector{SymbolicUtils.Code.Assignment}:
 SymbolicUtils.Code.Assignment(var"##cse#235", t^2)
 SymbolicUtils.Code.Assignment(var"##cse#236", 1 + t + var"##cse#235")
 SymbolicUtils.Code.Assignment(var"##cse#237", dde(var"##cse#236"))
 SymbolicUtils.Code.Assignment(var"##cse#238", sin(var"##cse#237"))
 SymbolicUtils.Code.Assignment(var"##cse#239", var"##cse#235"*var"##cse#237"*var"##cse#238")
```

AayushSabharwal (approved):
I can't wait to use this in MTK

src/code.jl
```diff
@@ -696,6 +697,46 @@ end
...
+        end
+        if iscall(node)
+            args = map(dfs, arguments(node))
+            new_node = maketerm(typeof(node), operation(node), args, metadata(node))
+            sym = newsym(symtype(new_node))
+            push!(sorted_nodes, sym ← new_node)
+            visited[node] = sym
+            return sym
+        else
+            visited[node] = node
```
AayushSabharwal, line 731:
```suggestion
            return sym
        elseif _is_array_of_symbolics(node)
            new_node = map(dfs, node)
            sym = newsym(typeof(new_node))
            push!(sorted_nodes, sym ← new_node)
            visited[node] = sym
            return sym
        else
            visited[node] = node
```

This enables it to also handle arrays of symbolics

AayushSabharwal (changes requested):
Turns out this doesn't handle arrays of symbolics. The fix is pretty straightforward.

Before:
```julia
julia> @syms a b c
julia> function foo end
julia> ex = term(foo, [a^2 + b^2, b^2 + c], c; type = Real)
julia> topological_sort(ex)
1-element Vector{SymbolicUtils.Code.Assignment}:
 SymbolicUtils.Code.Assignment(var"##cse#230", foo(SymbolicUtils.BasicSymbolic{Number}[a^2 + b^2, c + b^2], c))
```

After:
```julia
julia> topological_sort(ex)
6-element Vector{SymbolicUtils.Code.Assignment}:
 SymbolicUtils.Code.Assignment(var"##cse#235", a^2)
 SymbolicUtils.Code.Assignment(var"##cse#236", b^2)
 SymbolicUtils.Code.Assignment(var"##cse#237", var"##cse#235" + var"##cse#236")
 SymbolicUtils.Code.Assignment(var"##cse#238", c + var"##cse#236")
 SymbolicUtils.Code.Assignment(var"##cse#239", SymbolicUtils.BasicSymbolic{Number}[var"##cse#237", var"##cse#238"])
 SymbolicUtils.Code.Assignment(var"##cse#240", foo(var"##cse#239", c))
```

Codegen also works fine with this:
```julia
julia> SymbolicUtils.Code.toexpr(cse(ex))
:(let var"##cse#247" = (^)(a, 2), var"##cse#248" = (^)(b, 2), var"##cse#249" = (+)(var"##cse#247", var"##cse#248"), var"##cse#250" = (+)(c, var"##cse#248"), var"##cse#251" = begin
              #= /Users/aayush/Julia/SciML/SymbolicUtils.jl/src/code.jl:484 =#
              (SymbolicUtils.Code.create_array)(Vector{SymbolicUtils.BasicSymbolic{Number}}, nothing, Val{1}(), Val{(2,)}(), var"##cse#249", var"##cse#250")
          end
      (foo)(var"##cse#251", c)
  end)
```

ChrisRackauckas (approved):

PR #689, opened 2025-01-09
Simplify `NaNMath.pow` substitution

devmotion:
Line 149 was wrong, but it seems it was not caught by the tests since this simplification is applied already when the expression is constructed. Which made me wonder whether we could remove this branch completely from the `NaNMath.pow` substitution.

### DIFF
--- a/src/code.jl
+++ b/src/code.jl
@@ -139,18 +139,8 @@ function function_to_expr(op::Union{typeof(*),typeof(+)}, O, st)
 end
 
 function function_to_expr(op::typeof(^), O, st)
-    args = arguments(O)
-    if length(args) == 2 && args[2] isa Real && args[2] < 0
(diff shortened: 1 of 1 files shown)


devmotion:
Initially I assumed it would, but based on the tests I ran it didn't, which is why I opted for removing this special branch completely. My hypothesis was that it is already covered by the existing rewriting/simplification performed for `^` at construction of the symbolic expression, as mentioned in https://docs.sciml.ai/SymbolicUtils/stable/#Simplification.

AayushSabharwal:
The `NaNMath.pow` case hits this, since while `a ^ (-2)` might transform into `1 / a ^ 2`, `NaNMath.pow(a, -2)` transforms into `a ^ (-2)` which needs the rewrite rule.

PR #692, opened 2025-01-13
test: update codegen tests

AayushSabharwal:


devmotion (approved):

test/code.jl
```diff
@@ -101,18 +101,18 @@ nanmath_st.rewrites[:nanmath] = true
...
-    @test toexpr(NaNMath.pow(a, -1)) == :($(NaNMath.pow)(a, -1))
-    @test toexpr(NaNMath.pow(a, -1), nanmath_st) == :($(NaNMath.pow)(a, -1))
+    @test toexpr(NaNMath.pow(a, -1)) == :($(inv)(a))
+    @test toexpr(NaNMath.pow(a, -1), nanmath_st) == :($(inv)(a))
 
     @test toexpr(a^-2) == :($(/)(1, $(^)(a, 2)))
     @test toexpr(a^-2, nanmath_st) == :($(/)(1, $(NaNMath.pow)(a, 2)))
-    @test toexpr(NaNMath.pow(a, -2)) == :($(NaNMath.pow)(a, -2))
-    @test toexpr(NaNMath.pow(a, -2), nanmath_st) == :($(NaNMath.pow)(a, -2))
+    @test toexpr(NaNMath.pow(a, -2)) == :($(NaNMath.pow)($(inv)(a), 2))
```
devmotion, line 114: Actually, why is this and the line below not
```suggestion
    @test toexpr(NaNMath.pow(a, -2)) == :($(^)($(inv)(a), 2))
```

devmotion, line 114: What happens in this case?

AayushSabharwal: I'll make it not do that if the exponent is an integer

src/code.jl
```diff
@@ -146,11 +146,13 @@ function function_to_expr(op::typeof(^), O, st)
             return toexpr(Term(inv, Any[ex]), st)
         else
             args = Any[Term(inv, Any[ex]), -args[2]]
-            op = get(st.rewrites, :nanmath, false) ? op : NaNMath.pow
+            op = get(st.rewrites, :nanmath, false) || args[2] isa Integer ? op : NaNMath.pow
```
devmotion, line 149: Do we need this at all, see https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/689? In any case, the existing check is wrong and should be
```suggestion
            op = get(st.rewrites, :nanmath, false) === true && !(args[2] isa Integer) ? NaNMath.pow : op
```

src/code.jl
```diff
@@ -146,11 +146,13 @@ function function_to_expr(op::typeof(^), O, st)
...
             return toexpr(Term(op, args), st)
         end
     end
-    get(st.rewrites, :nanmath, false) === true || return nothing
+    if !get(st.rewrites, :nanmath, false) || args[2] isa Integer
```
devmotion, line 153: A bit safer:
```suggestion
    if get(st.rewrites, :nanmath, false) !== true || args[2] isa Integer
```
Additionally, I guess it would be more efficient to move this to the beginning of the function body?

PR #697, opened 2025-01-27
feat: cache hash2 in BasicSymbolic

AayushSabharwal:


AayushSabharwal:
Tests passed locally with https://github.com/JuliaMath/NaNMath.jl/pull/90

PR #704, opened 2025-02-13
Refactor `BasicSymbolic` struct to separate metadata from hash consing

bowenszhu:
This PR changes the structure of `BasicSymbolic` while avoid breaking changes.
| Old | New |
| ------------- | ------------- |
| current | next |

More detailed description will be added later.


src/types.jl
```diff
@@ -169,7 +194,10 @@ end
 TermInterface.children(x::BasicSymbolic) = arguments(x)
 TermInterface.sorted_children(x::BasicSymbolic) = sorted_arguments(x)
 function TermInterface.arguments(x::BasicSymbolic)
-    @compactified x::BasicSymbolic begin
+    arguments(x.expr)
```
AayushSabharwal, line 197: This should take `arguments(x.expr)` and wrap them in a `BasicSymbolic` with the corresponding metadata from `x.meta`, otherwise this breaks existing code. `metadata_children` would also not be necessary.

bowenszhu: OK. Thank you

PR #707, opened 2025-03-01
feat: add `@cache` macro

### LINKED ISSUE #705: Thread-specific hashconsing cache?
It struck me that we have thread-local RNG state in base, and might be able to use something similar here for our global hashconsing state. I don't quite understand what we do for the RNG right now, but the earlier solution [here](https://github.com/JuliaLang/julia/pull/32407/files) might be pretty easily applicable.

AayushSabharwal:
Close #705


src/cache.jl
```diff
@@ -0,0 +1,359 @@
...
+    argexprs = []
+
+    for arg in fn.args
+        # handle arguments with defaults
+        if Meta.isexpr(arg, :kw)
+            arg = arg.args[1]
+        end
+        if !Meta.isexpr(arg, :(::))
+            # if the type is `Any`, branch on it being a `BasicSymbolic`
+            conditions = :($conditions && (!($arg isa BasicSymbolic) || $arg.uuid !== nothing))
```
bowenszhu, line 229: The `uuid` field was removed. So the condition here should be modified.

src/cache.jl
```diff
@@ -0,0 +1,359 @@
...
+        argname, Texpr = arg.args
+        push!(argexprs, argname)
+
+        # handle Union types that may contain a `BasicSymbolic`
+        if Meta.isexpr(Texpr, :curly) && Texpr.args[1] == :Union
+            Texprs = Texpr.args[2:end]
+            Ts = map(Base.Fix1(Base.eval, __module__), Texprs)
+            keyTs = map(x -> x <: BasicSymbolic ? UInt64 : x, Ts)
+            if any(x -> x <: BasicSymbolic, Ts)
+                conditions = :($conditions && (!($argname isa BasicSymbolic) || $argname.uuid !== nothing))
```
bowenszhu, line 244: `uuid`

bowenszhu (commented):
This adds thread-safe and fine-grained control over cached function behavior which is expected to enhances performance for recursive symbolic computations.

Could tests be added for the following?
- cache hits/misses for varying subtypes of `BasicSymbolic`
- cache eviction when the `limit` is reached
- task-local behavior
- thread safety

bowenszhu (approved):
Nice memoization implementation.

AayushSabharwal:
🎉 thanks! I'll rebase

PR #709, opened 2025-03-10
fix: handle explicit `::Any` annotation in `@cache` macro

AayushSabharwal:
Close https://github.com/JuliaSymbolics/Symbolics.jl/issues/1470
Close https://github.com/SciML/ModelingToolkit.jl/issues/3447

### DIFF
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -245,6 +245,13 @@ macro cache(args...)
         argname, Texpr = arg.args
         push!(argexprs, argname)
 
+        if Texpr == :Any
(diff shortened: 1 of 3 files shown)


AayushSabharwal:
This should fix things. A better solution would likely be a `struct CacheKey{T <: Tuple}; x::T; end` which has `@generated` impls for `hash` and `isequal` that use `objectid`/`===` for `BasicSymbolic` the way `@cache` does and falls back for other types. I had already implemented this by the time that discussion came up, and since this is a big failure it seems prudent to fix fast and clean it later.

PR #718, opened 2025-03-19
refactor: disable internal mutation of `BasicSymbolic`

AayushSabharwal:


AayushSabharwal:
This is nice but breaks Symbolics precompilation, and as such I think should be part of a breaking version that includes the shift to Moshi.

PR #721, opened 2025-03-23
fix: remove Metatheory from benchmark environment

AayushSabharwal:


AayushSabharwal:
Seriously? Why is CSE stochastic now.

AayushSabharwal:
I think the test needs changing. This seems to be GC-related. Whenever it reproduces locally (which isn't often) it corresponds to a cache miss.

PR #723, opened 2025-03-23
feat: optimize small arrays in `BasicSymbolic`

AayushSabharwal:


AayushSabharwal:
I like those performance numbers

AayushSabharwal:
I'll setup downstream benchmarks sometime soon, so we can get better numbers on actual use cases while still being able to run the core benchmarks on a breaking release.

src/small_array.jl
```diff
@@ -0,0 +1,133 @@
...
+    end
+end
+
+function Base.push!(x::Backing, v)
+    x.len < 3 || throw(ArgumentError("`Backing` is full"))
+    x.len += 1
+    x[x.len] = v
+end
+
+function Base.pop!(x::Backing{T}, v) where {T}
```
bowenszhu, line 63:
```suggestion
function Base.pop!(x::Backing{T}) where {T}
```
The 2nd argument does not seem to be necessary.

AayushSabharwal: Thanks!

PR #726, opened 2025-03-24
refactor: return to using `WeakValueDict`

AayushSabharwal:
`WeakKeyDict` didn't actually do anything. The thing being hashconsed was `HashconsingWrapper`, which doesn't have a reference outside of the cache. So it was basically up to the GC how much hashconsing we do.


AayushSabharwal:
Basically, the symbolic expression has references to the `BasicSymbolic`, so it isn't garbage collected. However, nothing retains a reference to the `HashconsingWrapper`, which means it was GCed.

bowenszhu (approved):

PR #727, opened 2025-03-25
fix: fix keyword syntax for `@cache` options

AayushSabharwal:


AayushSabharwal:
n-ary addition causes problems when adding with `Num`s, since they subtype real. This is incredibly annoying because that is a significant source of allocations.

PR #731, opened 2025-03-27
name CSE states in a reproducible way

hexaeder:
The goals of this PR is make sure, that CSE variables are named in a reproducible way.
Otherwise, built functions are not considered egal which can lead to higher memory usage and hinders type stability in some contexts.
```julia
using Symbolics
@variables x, y, z
formulas = [x, x+y, x+y+z]
f1 = Symbolics.build_function(formulas, [x, y, z]; expression=false, cse=true)
f2 = Symbolics.build_function(formulas, [x, y, z]; expression=false, cse=true)
f1===f2
```

### DIFF
--- a/src/code.jl
+++ b/src/code.jl
@@ -737,7 +737,16 @@ end
 
 ### Common subexprssion evaluation
 
-@inline newsym(::Type{T}) where T = Sym{T}(gensym("cse"))
+"""
+    newsym!(state::CSEState, ::Type{T})
+
+Generates new symbol of type `T` with unique name in `state`.
+"""
+@inline function newsym!(state, ::Type{T}) where T 
+    name = "##cse#$(state.varid[])"
+    state.varid[] += 1
+    Sym{T}(Symbol(name))
+end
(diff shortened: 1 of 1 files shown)


src/code.jl
```diff
@@ -737,7 +737,16 @@ end
...
+"""
+    newsym!(state::CSEState, ::Type{T})
+
+Generates new symbol of type `T` with unique name in `state`.
+"""
+@inline function newsym!(state, ::Type{T}) where T 
+    name = "##cse#$(state.varid[])"
+    state.varid[] += 1
+    Sym{T}(Symbol(name))
+end
```
bowenszhu, line 749: What if a user maliciously creates a `Sym` with name `##cse#1`?

AayushSabharwal: We're not really building a secure software. There's already a lot of ways users can make our stuff break if they look into internals, so it doesn't feel worth to try and guard against this.

bowenszhu (approved):

PR #732, opened 2025-03-28
Refactor get_degrees() to make it faster

alyst:
Profiling indicated that in my use case (symbolic multiplication of several large sparse matrices takes almost 1h) a lot of time is spent in `get_degrees()`.

So this PR is my attempt to make it faster and use less memory:
  * some trivial fixes to avoid broadcasting
  * extract operator handling into separate routines (`get_degrees(op, expr)`) instead of `if op == + .. elseif op == * end`.
    That increases memory footprint by 10%, but also makes it faster by 10% (I guess type inference is improved).
  * avoid calling `sorted_arguments()` within `get_degrees()`, as that itself calls `get_degrees()`, so `get_degrees()` gets called multiple times for the same expression.
    Instead, `get_degrees()` uses `arguments()` and sorts the list of degrees itself.
  * Cache the `get_degrees(getindex, expr)`: that increases the speed and 2x reduces memory footprint (in the test example below)

On the test case I observe 3x speed-up and 3x reduction in memory footprint.

The benchmark script
```julia

@syms x y[1:3] z[1:2, 1:2]
# Create a complex polynomial expression
y1 = term(getindex, y, 1, type=Number)
y2 = term(getindex, y, 2, type=Number)
y3 = term(getindex, y, 3, type=Number)
z11 = term(getindex, z, 1, 1, type=Number)
z12 = term(getindex, z, 1, 2, type=Number)
z23 = term(getindex, z, 2, 3, type=Number)
large_poly = SymbolicUtils.expand((x^2 + 2y1 + 3z12 + y2*z23 + x*y1*z12 - x^2*z12 + x*z11 + y3 + y2 + z23 + 1)^8);

b = @benchmark SymbolicUtils.get_degrees($large_poly) samples=50 evals=1
```

Results on a *master* (Julia 1.11.4):
```
BenchmarkTools.Trial: 16 samples with 1 evaluation per sample.
 Range (min … max):  298.827 ms … 359.108 ms  ┊ GC (min … max): 0.00% … 12.83%
 Time  (median):     311.555 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   319.948 ms ±  19.761 ms  ┊ GC (mean ± σ):  4.18% ±  5.27%

      █                                   ▁                      
  ▆▆▁▆█▁▁▆▁▁▁▁▁▁▁▁▁▁▆▁▁▁▁▁▁▁▁▁▁▁▁▁▆▁▆▁▁▁▆▁█▁▁▁▁▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆ ▁
  299 ms           Histogram: frequency by time          359 ms <

 Memory estimate: 116.91 MiB, allocs estimate: 3390531.

BenchmarkTools.Trial: 16 samples with 1 evaluation per sample.
 Range (min … max):  299.432 ms … 360.751 ms  ┊ GC (min … max): 0.00% … 13.12%
 Time  (median):     318.398 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   320.586 ms ±  19.074 ms  ┊ GC (mean ± σ):  4.72% ±  5.29%

  ▃   ▃█                                                         
  █▁▁▇██▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▇▇▇▁▇▇▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇ ▁
  299 ms           Histogram: frequency by time          361 ms <

 Memory estimate: 116.91 MiB, allocs estimate: 3390531.
```

Results for this PR:
```
BenchmarkTools.Trial: 43 samples with 1 evaluation per sample.
 Range (min … max):  107.692 ms … 156.781 ms  ┊ GC (min … max): 0.00% … 22.67%
 Time  (median):     109.903 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   116.491 ms ±  11.735 ms  ┊ GC (mean ± σ):  3.87% ±  6.74%

  █▁                                                             
  ██▃▁▁▃▁▃▃▁▄▁▃▁▁▁▆▄▁▄▃▁▁▃▁▁▁▁▁▃▁▃▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▃ ▁
  108 ms           Histogram: frequency by time          157 ms <

 Memory estimate: 43.92 MiB, allocs estimate: 1448421.

BenchmarkTools.Trial: 43 samples with 1 evaluation per sample.
 Range (min … max):  107.790 ms … 146.127 ms  ┊ GC (min … max): 0.00% … 23.08%
 Time  (median):     111.730 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   116.886 ms ±  10.678 ms  ┊ GC (mean ± σ):  3.84% ±  6.82%

  █                                                              
  ██▆▃▁▁▃▁▃▁▃▁▁▃▃▁▁▁▁▁▃▄▃▇▁▆▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▃▁▁▁▃▁▁▁▁▁▃▁▁▃ ▁
  108 ms           Histogram: frequency by time          146 ms <

 Memory estimate: 43.92 MiB, allocs estimate: 1448421.
```

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -5,7 +5,7 @@ using Random
 
 SUITE = BenchmarkGroup()
 
-@syms a b c d; Random.seed!(123);
+@syms a b c d x y[1:3] z[1:2, 1:2]; Random.seed!(123);
 
 let r = @rule(~x => ~x), rs = RuleSet([r]),
     acr = @rule(~x::is_literal_number + ~y => ~y)
@@ -67,7 +67,19 @@ let r = @rule(~x => ~x), rs = RuleSet([r]),
         subs_expr = (sin(a+b) + cos(b+c)) * (sin(b+c) + cos(c+a)) * (sin(c+a) + cos(a+b))
     end
 
+    overhead["get_degrees"] = BenchmarkGroup()
 
+    let y1 = term(getindex, y, 1, type=Number),
+        y2 = term(getindex, y, 2, type=Number),
+        y3 = term(getindex, y, 3, type=Number),
+        z11 = term(getindex, z, 1, 1, type=Number),
+        z12 = term(getindex, z, 1, 2, type=Number),
+        z23 = term(getindex, z, 2, 3, type=Number),
+        # create a relatively large polynomial
+        large_poly = SymbolicUtils.expand((x^2 + 2y1 + 3z12 + y2*z23 + x*y1*z12 - x^2*z12 + x*z11 + y3 + y2 + z23 + 1)^8)
+
+        overhead["get_degrees"]["large_poly"] = @benchmarkable SymbolicUtils.get_degrees($large_poly)
+    end
 end
 
 let
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -74,10 +74,10 @@ let r = @rule(~x => ~x), rs = RuleSet([r]),
         y3 = term(getindex, y, 3, type=Number),
         z11 = term(getindex, z, 1, 1, type=Number),
         z12 = term(getindex, z, 1, 2, type=Number),
-        z23 = term(getindex, z, 2, 3, type=Number),
+        z23 = term(getindex, z, 2, 3, type=Number)
+
         # create a relatively large polynomial
         large_poly = SymbolicUtils.expand((x^2 + 2y1 + 3z12 + y2*z23 + x*y1*z12 - x^2*z12 + x*z11 + y3 + y2 + z23 + 1)^8)
-
         overhead["get_degrees"]["large_poly"] = @benchmarkable SymbolicUtils.get_degrees($large_poly)
     end
 end
--- a/src/ordering.jl
+++ b/src/ordering.jl
@@ -30,7 +30,8 @@ For a sum expression, returns the `get_degrees()` result for term with the highe
 See also `monomial_lt` and `lexlt`.
 """
 function get_degrees(expr)
-    res = _get_degrees(expr)
+    degs_cache = Dict()
+    res = _get_degrees(expr, degs_cache)
     if res isa AbstractVector
         return Tuple(res)
     else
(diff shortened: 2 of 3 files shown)


AayushSabharwal:
Thanks for implementing this, the performance improvement is great. I have a couple concerns:
- It seems to have different behavior. The Symbolics reference test is failing
- Could `_get_degrees` be written so it doesn't dispatch on `op`? Since `BasicSymbolic` is type-unstable (out of necessity) it's always going to face runtime dispatch this way, and would be better off as an ifelse chain.

alyst:
> It seems to have different behavior. The Symbolics reference test is failing

Yes, I think that's because I have added support for `/`,  so now it puts the negative powers first.
I believe before fractions were going the "zzzz" + hash() path.
I can fix it, but probably that's the right moment to clarify what is the expected sorting for the terms with negative powers?
Should the terms be sorted by `abs(deg)`?

> Could _get_degrees be written so it doesn't dispatch on op? Since BasicSymbolic is type-unstable (out of necessity) it's always going to face runtime dispatch this way, and would be better off as an ifelse chain.

I can do that, but, as I wrote in the description, the dispatching version is 10% faster.
I guess that's because type inference for local variables that are reused by the if branches works better when the code is in separate functions.
So maybe if the variable names in different op branches don't overlap, it will be faster -- but if that's the case, it is some fragile code with obscure requirements to be fast (plus it is less readable).

alyst:
I've changed the sorting of negative degree:
 * for `lexlt()`: terms are sorted by descending *abs(deg)*, if equal, positive degree is first
 * for `monomial_lt()`: terms are sorted by ascending *abs(deg)*, if equal, positive degree is first

That should fix the *Symbolics.jl* integration tests (I've also added the failing expression to *SymbolicUtils.jl* unit tests).
The failures in *ModelingToolkit.jl* don't seem to be directly related to this PR. Probably some spurious numerical method failures.

alyst:
@AayushSabharwal I have updated the term sorting logic to fix the integration tests. Could you please re-run them?

AayushSabharwal:
This looks good. I _think_ the Catalyst failure is possibly `structural_simplify` being temperamental. Now that benchmarks work, could you add your script above to the benchmark CI?

AayushSabharwal:
Could you also rebase this on top of master?

alyst:
All tests seem to pass

AayushSabharwal:
Why does the benchmark refuse to run the one you added?

alyst:
@copilot Why the new benchmarks are not run?

alyst:
@AayushSabharwal Sorry for the long delay. Maybe it has something to do with `large_poly` var being new variable binding for the let block. I have changed that in the last commit.
But I suspect the real reason is that the PR is checked against the benchmarks defined in the default branch, not the PR itself, see how [benchpkg is run](https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/master/.github/workflows/benchmark_pr.yml#L41).
So for the `"get_degrees"` benchmark to show up in the PR reports, you need to accept this PR first.

I have checked that locally "get_degrees" will run as a part of `SUITE`:
```
julia> results = run(SUITE)
2-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "polyform" => 4-element BenchmarkTools.BenchmarkGroup:
          tags: []
          "easy_iszero" => Trial(35.500 μs)
          "simplify_fractions" => Trial(1.696 ms)
          "iszero" => Trial(1.240 ms)
          "isone" => Trial(0.001 ns)
  "overhead" => 7-element BenchmarkTools.BenchmarkGroup:
          tags: []
          "ruleset" => 3-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "noop:Int" => Trial(100.000 ns)
                  "noop:Sym" => Trial(100.000 ns)
                  "noop:Term" => Trial(4.000 μs)
          "simplify" => 7-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "randterm (/, *):serial" => Trial(211.300 μs)
                  "noop:Int" => Trial(200.000 ns)
                  "noop:Sym" => Trial(200.000 ns)
                  "randterm (/, *):thread" => Trial(248.300 μs)
                  "noop:Term" => Trial(45.400 μs)
                  "randterm (+, *):serial" => Trial(127.347 ms)
                  "randterm (+, *):thread" => Trial(149.125 ms)
          "acrule" => 5-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "noop:Int" => Trial(0.001 ns)
                  "noop:Sym" => Trial(0.001 ns)
                  "a+2+b" => Trial(800.000 ns)
                  "a+b" => Trial(200.000 ns)
                  "a+2" => Trial(800.000 ns)
          "large_poly" => 0-element BenchmarkTools.BenchmarkGroup:
                  tags: []
          "rule" => 3-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "noop:Int" => Trial(0.001 ns)
                  "noop:Sym" => Trial(0.001 ns)
                  "noop:Term" => Trial(0.001 ns)
          "substitute" => 3-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "a,b,c" => Trial(15.900 μs)
                  "a,b" => Trial(73.700 μs)
                  "a" => Trial(85.700 μs)
          "get_degrees" => 1-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "large_poly" => Trial(141.879 ms)

julia> results["overhead"]["get_degrees"]["large_poly"]
BenchmarkTools.Trial: 32 samples with 1 evaluation per sample.
 Range (min … max):  141.879 ms … 221.265 ms  ┊ GC (min … max): 0.00% … 17.56%
 Time  (median):     157.718 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   159.829 ms ±  16.505 ms  ┊ GC (mean ± σ):  2.88% ±  5.48%

    ██       ▄  ▁ ▁    ▁
  ▆▆██▆▆▁▁▆▁▁█▆▁█▆█▆▆▆▆█▁▁▁▁▆▁▁▁▁▁▁▆▁▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆ ▁
  142 ms           Histogram: frequency by time          221 ms <

 Memory estimate: 43.92 MiB, allocs estimate: 1448503.
```

PR #749, opened 2025-06-03
[WIP] first prototype of default value rules implemented

Bumblebee00:
Reading Mathematica pattern matching documantation ([here](https://reference.wolfram.com/language/tutorial/Patterns.html)) I saw they have a **pattern matching with slots that can take a default value**. In Mathematica a generic pattern named "a" is written a_  (in julia symbolics ~a ). But then if I write a_.  it matches a pattern with built in default value, so basically (a_. + b_*x_)  can match 2x where "a" takes the default value of 0 (and b=2 of course). or another example (a_ + b_*x_)^m_.  can match (1 + 2x) where m takes the default value of 1 (and a=1 b=2 of course), whereas the pattern  (a_ + b_*x_)^m_  would not match (1 + 2x) because the exponent m needs to be there and cannot take a default value.

This feature is needed for my gsoc project of implementing a rule based integrator in julia symbolics.

Now I implemented this default value for sum and multiplication.
I choose the syntax of this new slot to be `~!x`, so for example:
```
r_mult = @rule (~x * ~!y + ~z) => ~y
r_mult(c + a*b) # returns b
r_mult(c + b) # returns 1
```
whereas
```
r_mult_old = @rule (~x * ~y + ~z) => ~y
r_mult_old(c + a*b) # returns b
r_mult_old(c + b) # returns nothing because it did not match
```

### DIFF
--- a/src/matchers.jl
+++ b/src/matchers.jl
@@ -5,25 +5,51 @@
 # 2. Dictionary
 # 3. Callback: takes arguments Dictionary × Number of elements matched
 #
+
 function matcher(val::Any)
-    iscall(val) && return term_matcher(val)
+    # if val is a call (like an operation) creates a term matcher or term matcher with defslot
+    if iscall(val)
+        # if has two arguments and one of them is a DefSlot, create a term matcher with defslot
+        if length(arguments(val)) == 2 && any(x -> isa(x, DefSlot), arguments(val))
+            return term_matcher_defslot(val)
+        # else return a normal term matcher
+        else
+            return term_matcher(val)
+        end
+    end
+
     function literal_matcher(next, data, bindings)
+        # car data is the first element of data
         islist(data) && isequal(car(data), val) ? next(bindings, 1) : nothing
     end
 end
 
 function matcher(slot::Slot)
     function slot_matcher(next, data, bindings)
-        !islist(data) && return
+        !islist(data) && return nothing
         val = get(bindings, slot.name, nothing)
+        # if slot name already is in bindings, check if it matches
         if val !== nothing
             if isequal(val, car(data))
                 return next(bindings, 1)
             end
-        else
-            if slot.predicate(car(data))
-                next(assoc(bindings, slot.name, car(data)), 1)
+        # elseif the first element of data matches the slot predicate, add it to bindings and call next
+        elseif slot.predicate(car(data))
+            next(assoc(bindings, slot.name, car(data)), 1)
+        end
+    end
+end
+
+function matcher(defslot::DefSlot)
+    function defslot_matcher(next, data, bindings)
+        !islist(data) && return
+        val = get(bindings, defslot.name, nothing)
+        if val !== nothing
+            if isequal(val, car(data))
+                return next(bindings, 1)
             end
+        elseif defslot.predicate(car(data))
+            next(assoc(bindings, defslot.name, car(data)), 1)
         end
     end
 end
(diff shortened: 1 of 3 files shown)


AayushSabharwal:
Nice! I'm a little concerned about ordering here. How does this handle commutative operations?

```julia
r_mult = @rule (~x * ~!y + ~z) => ~y
r_mult(c*d + a*b) # returns?
```

This could reasonably return any one of `a`, `b`, `c`, or `d`.

```julia
r_mult(2c + b) # returns?
```

This could return 1 or 2.

How are these cases handled?

Bumblebee00:
@AayushSabharwal the symbols are reorderd in a kind of alphabetical order before being passed to the rule, we can also check it from julia repl:
```julia
julia> c*d + a*b
a*b + c*d

julia> 2c + b
b + 2c
```
being the rule macro not commutiative, is returned the vaule that matches the rule in the ordered expression, so:
```julia
julia> r_mult(c*d + a*b) # is like r_mult(a*b + c*d)
b

julia> r_mult(2c + b) # is like r_mult(b + 2c)
1
```
or, another example:
```julia
julia> r_mult1 = @rule (~z + ~!y * ~x) => ~y
~z + ~(!y) * ~x => ~y

julia> r_mult2 = @rule (~z + ~x * ~!y) => ~y
~z + ~x * ~(!y) => ~y

julia> r_mult1(1+a*b)
a

julia> r_mult2(1+a*b)
b

julia> r_mult1(1+a)
1

julia> r_mult2(1+a)
1
```



I didn't think about this, but maybe for the symbolic integrator I need [acrules](https://symbolicutils.juliasymbolics.org/rewrite/#associative-commutative_rules)... I dont know

AayushSabharwal:
> the symbols are reorderd in a kind of alphabetical order before being passed to the rule, we can also check it from julia repl

The printout in the REPL can be misleading. It uses `sorted_arguments` to print, which in general (especially for addition/multiplication) is not the same as `arguments`. The order is also not purely lexicographic, it uses a heuristic based on the degree of each term.

> I didn't think about this, but maybe for the symbolic integrator I need [acrules](https://symbolicutils.juliasymbolics.org/rewrite/#associative-commutative_rules)

ACRules are likely necessary for some integration rules, but it doesn't help with the ambiguity. That said, I'm not sure if the ambiguity actually matters for integration. It seems all our existing matchers use `arguments` and rely on whatever ordering comes out, so I guess the ambiguity is fine from a relative robustness standpoint.

Bumblebee00:
> The printout in the REPL can be misleading
Yes you are right, in fact after writing the message I found some cases I could not explain:
```julia
julia> r_mult1(1+c*d) # I expect c but
d

julia> r_mult2(1+c*d) # I expect d but
c
```
and in fact:
```julia
julia> sorted_arguments(c*d)
2-element Vector{Any}:
 c
 d

julia> arguments(c*d)
2-element SymbolicUtils.SmallVec{Any, Vector{Any}}:
 d
 c
```

src/matchers.jl
```diff
@@ -140,25 +133,43 @@ function term_matcher(term)
...
+    matchers = (matcher(operation(term)), map(matcher, a)...) # create matchers for the operation and the two arguments of the term
+    
+    defslot_index = findfirst(x -> isa(x, DefSlot), a) # find the defslot in the term
+    defslot = a[defslot_index]
+    
+    function defslot_term_matcher(success, data, bindings)
+        # if data is not a list, return nothing
+        !islist(data) && return nothing
+        # if data (is not a tree and is just a symbol) or (is a tree not starting with the default operation)
+        if !iscall(car(data)) || (istree(car(data)) && string(defslot.operation) != string(operation(car(data))))
```
AayushSabharwal, line 159: Why does this use string comparison?

src/rule.jl
```diff
@@ -31,53 +31,61 @@ function makeslot(s::Expr, keys)
...
+function defaultValOfCall(call)
+    if call == :+
+        return 0
+    elseif call == :*
+        return 1
+    elseif call == :^
+        return 1
+    end
+    # else no default value for this call
+    return nothing
```
AayushSabharwal, line 67: Why return `nothing` instead of erroring?

src/matchers.jl
AayushSabharwal: Ah, https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/749#issuecomment-2939417395 didn't load for me until I refreshed the page. That behavior is... weird. However, you should be able to match `nameof(operation(...))` with `defslot.operation`.

Bumblebee00: Thanks I didn't know of nameof, I will use it now

src/rule.jl
Bumblebee00: Yeah I put it there at the beginning and forgot, is better to trow an error. I will add it now

PR #750, opened 2025-06-14
[WIP] Negative exponent matching

Bumblebee00:
This pr builds on top of the defslot feature i added in another pr. The aim is for rule `@rule (~x)^(~m) => ..something..` to match 1 / (x^2) with m=-2.

I did that by modifying the `term_matcher` and `defslot_term_matcher` functions, they are functions called when the rule gets applied, and they have to check wheter a certain term (the pattern to match in the ruel, like in this example `(~x)^(~m)`) matches with certain data (1/(x^2) in this example). I did basically the same modification to the two of them. They work by using a loop function called recursively, that uses some matcher functions (1) that assign the values to their label. The modification di did in `term_matcher` is:
- call the loop function recursively like normal.
- If still you dont have a match, and if the operation of the term is a power, and if data is in the form 1 over a power, like 1/a^b, then:
- - if the exponent of the power b is not a number (but a symbol or expression), call the loop function with a^-b
- - if b is a number, this cannot be done bc x^-3 gets [automatically transformed](https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/a95aea2310ea12f1ee2d9b2850d062ec82a437c4/src/types.jl#L1535-L1550) into 1 / (x^3). So I cant call loop with a^-b, i have to call it with a^b, so I had to create a new matcher function (1) that assigns the opposite value of data to the label.

in `defslot_term_matcher` i did a similar thing, and if after all this still doesnt find a match, also tries to match with the default value of defslot

I find it a bit messy

### DIFF
--- a/src/matchers.jl
+++ b/src/matchers.jl
@@ -13,9 +13,6 @@ function matcher(val::Any)
         # just two arguments bc defslot is only supported with operations with two args: *, ^, +
         if length(arguments(val)) == 2 && any(x -> isa(x, DefSlot), arguments(val))
             return defslot_term_matcher_constructor(val)
-        # else (a)^(b) can also match 1/( (a)^(b) ) , just with b of oppsite sign
-        elseif operation(val) == ^
-            return neg_pow_term_matcher_constructor(val)
         # else return a normal term matcher
         else
             return term_matcher_constructor(val)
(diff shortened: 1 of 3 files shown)


Bumblebee00:
moved to https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/752

PR #719, opened 2025-03-19
Fix for complex powers of symbolics

karlwessel:
This fixes issue #717

Issue #717, opened 2025-03-19
Power with complex exponent fails


karlwessel:
Taking the power of a symbolic and a complex exponent fails:
```julia repl
julia> using SymbolicUtils

julia> @syms x
(x,)

julia> x^im
ERROR: MethodError: no method matching isless(::Complex{Bool}, ::Int64)
The function `isless` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  isless(::Missing, ::Any)
   @ Base missing.jl:87
  isless(::Any, ::Missing)
   @ Base missing.jl:88
  isless(::AbstractFloat, ::Real)
   @ Base operators.jl:180
  ...

Stacktrace:
 [1] <(x::Complex{Bool}, y::Int64)
   @ Base ./operators.jl:353
 [2] ^(a::SymbolicUtils.BasicSymbolic{Number}, b::Complex{Bool})
   @ SymbolicUtils ~/.julia/packages/SymbolicUtils/QUEf3/src/types.jl:1493
 [3] top-level scope
   @ REPL[5]:1
```
 I'll submit a PR with a fix.

PR #751, opened 2025-06-15
fix issue #717

karlwessel:
Continuation of accidentally closed https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/719.


karlwessel:
With a rebase only two of the Benchmarks are significantly slower now.

PR #752, opened 2025-06-18
[WIP] Commutative operations and negative exponent match in rules

Bumblebee00:
this pr builds on top of [defslot](https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/749) and extends [negative exponent matching](https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/750). Before this pr julia symbolic rewriting rules had this unfortunate behaviour:
```
julia> @syms a x
(a, x)

julia> rule = @rule sin(~x) + cos(~x) => "yes"
sin(~x) + cos(~x) => 1

julia> arguments(sin(x) + cos(x))
2-element SymbolicUtils.SmallVec{Any, Vector{Any}}:
 cos(x)
 sin(x)

julia> rule(sin(x)+cos(x)) # doesnt match bc arguments are ordered cos first and then sin


julia> arguments(sin(a) + cos(a))
2-element SymbolicUtils.SmallVec{Any, Vector{Any}}:
 sin(a)
 cos(a)

julia> rule(sin(a)+cos(a)) # match bc arguments are ordered sin first and then cos
"yes"
```
there was the [acrule](https://symbolicutils.juliasymbolics.org/rewrite/#associative-commutative_rules) macro that helped, but was working only on the first operation of the expression, not every level. with this pr instead every * and + operation creates a matcher that checks all possible orderings of its arguments. 

This can also be a bad thing becuse the cases to check could be too many...

Also I refacored the code from my last two pr to make everything clearer and cleaner

### DIFF
--- a/docs/src/manual/rewrite.md
+++ b/docs/src/manual/rewrite.md
@@ -180,13 +180,13 @@ It works. This can be further simplified using Pythagorean identity and check it
 ```jldoctest rewriteex
 pyid = @rule sin(~x)^2 + cos(~x)^2 => 1
 
-pyid(cos(x)^2 + sin(x)^2) === nothing
+pyid(sin(x)^2 + 2sin(x)*cos(x) + cos(x)^2)===nothing
 
 # output
 true
 ```
 
-Why does it return `nothing`? If we look at the rule, we see that the order of `sin(x)` and `cos(x)` is different. Therefore, in order to work, the rule needs to be associative-commutative.
+Why does it return `nothing`? If we look at the expression, we see that we have an additional addend `+ 2sin(x)*cos(x)`. Therefore, in order to work, the rule needs to be associative-commutative.
 
 ```jldoctest rewriteex
 acpyid = @acrule sin(~x)^2 + cos(~x)^2 => 1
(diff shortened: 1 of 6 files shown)


Bumblebee00:
To sum up, this pr includes:
- Associative commutative checks
- Negative exponent matching
- Matching of exp as e^x and of sqrt as x^1//2
- Some bug fixing and refactoring of the defslot feature

Bumblebee00:
To sum up this PR brings the following changes:
- Commutative checks on `+`, `*`: every possible combination of the arguments of a `+` and `*` opeation are tried
- Negative exponent matching: `(~x)^(~m)`matches `1/x^2` with `~m = -2`

Also, not related to negative exponents but:
- `@rule exp(~x)` matches `e^x` and `@rule e^(~x)` matches `exp(x)`
- `@rule sqrt(~x)` matches `x^(1//2)` and `@rule (~x)^(1//2)` matches `sqrt(x)`

Bumblebee00:
The failing thests are the ones that @AayushSabharwal told me not to worry about, so I think this pr is ready to merge

AayushSabharwal:
Thanks! Pretty complicated PR, but it's nice to have better matching.

PR #771, opened 2025-08-07
feat: add `Polyform` variant, remove `AddOrMul`, `Pow`

AayushSabharwal:
We can remove `Pow` because `SmallVec` doesn't allocate for fewer than 3 elements.


AayushSabharwal:
Still need to
- essentially nuke `polyform.jl`
- refactor and fix tests

PR #784, opened 2025-08-29
refactor: move some code around

AayushSabharwal:


AayushSabharwal:
Proof:

```julia
julia> @syms a b
(a, b)

julia> @benchmark +($a, $b)
[ Info: Loading BenchmarkTools ...
BenchmarkTools.Trial: 10000 samples with 54 evaluations per sample.
 Range (min … max):  863.426 ns … 81.327 μs  ┊ GC (min … max): 0.00% … 98.24%
 Time  (median):     936.722 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   976.475 ns ±  1.304 μs  ┊ GC (mean ± σ):  2.90% ±  2.18%

       ▃▇█▇▇█▇▆▂
  ▂▃▅▄▆██████████▆▅▄▄▄▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂ ▃
  863 ns          Histogram: frequency by time          1.3 μs <

 Memory estimate: 928 bytes, allocs estimate: 24.
```

On `master`

PR #789, opened 2025-09-04
feat: support symbolic arrays

AayushSabharwal:


AayushSabharwal:
This is just the basics, I'll make a separate PR for `ArrayOp` to avoid stretching this one.

AayushSabharwal:
Specifically handling arrays of symbolics is just purely extra work and I can't get multiplication closer to what it was before. It's not a huge deal - 9x faster than master instead of 14x

PR #809, opened 2025-11-07
fix: fix linear indexing of `StableIndex`, add more utility to it

AayushSabharwal:


AayushSabharwal:
The failing test in Symbolics is due to the test not obeying the new rules for called functions vs dependent variables.

PR #817, opened 2025-11-28
refactor: improve printing code, allow custom `show_call` methods

AayushSabharwal:


AayushSabharwal:
I need custom `show_call` methods for symbolic structs, so just made the code a bit better too. It _should_ be faster and more type-stable. I haven't yet tested whether it precompiles. Will add documentation for custom printing in a future PR once I'm sure of the contracts.

PR #829, opened 2025-12-11
Add `search_variables` method for `Let`

DhairyaLGandhi:


src/code.jl
```diff
@@ -1245,4 +1245,12 @@ function apply_optimization_rules(expr, state::Code.CSEState, rules)
     return nothing
 end
 
+function search_variables!(buf, expr::Let)
+    rhs_buf = Set()
+    lhs_buf = Set()
```
AayushSabharwal:
```suggestion
    rhs_buf = empty(buf)
    lhs_buf = empty(buf)
```

src/code.jl
```diff
@@ -1245,4 +1245,12 @@ function apply_optimization_rules(expr, state::Code.CSEState, rules)
...
+    search_variables!.(Ref(rhs_buf), rhs.(expr.pairs))
+    search_variables!.(Ref(lhs_buf), lhs.(expr.pairs))
+    union!(buf, setdiff(rhs_buf, lhs_buf))
```
AayushSabharwal:
```suggestion
    union!(buf, setdiff!(rhs_buf, lhs_buf))
```

src/code.jl
```diff
@@ -1245,4 +1245,12 @@ function apply_optimization_rules(expr, state::Code.CSEState, rules)
     return nothing
 end
 
+function search_variables!(buf, expr::Let)
+    rhs_buf = empty(buf)
+    lhs_buf = empty(buf)
+    search_variables!.(Ref(rhs_buf), rhs.(expr.pairs))
+    search_variables!.(Ref(lhs_buf), lhs.(expr.pairs))
+    union!(buf, setdiff!(rhs_buf, lhs_buf))
```
DhairyaLGandhi: only issue i have with this is that the return semantics aren't well defined. We aren't guaranteed to have the `rhs_buf` populated returned as opposed to `nothing`

src/code.jl
```diff
@@ -1245,4 +1245,12 @@ function apply_optimization_rules(expr, state::Code.CSEState, rules)
     return nothing
 end
 
+function search_variables!(buf, expr::Let)
+    rhs_buf = Set()
+    lhs_buf = Set()
+    search_variables!.(Ref(rhs_buf), rhs.(expr.pairs))
+    search_variables!.(Ref(lhs_buf), lhs.(expr.pairs))
```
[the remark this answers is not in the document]
AayushSabharwal: I meant that the broadcast will allocate unnecessarily. Couldn't this just be a `for` loop?

src/code.jl
AayushSabharwal: Oh, I see. You're referring to return semantics of `setdiff!` - I assumed you were talking about `search_variables!` since this is a separate comment. Yeah, if you want this can be phrased as
```julia
setdiff!(rhs_buf, lhs_buf)
union!(buf, rhs_buf)
```

PR #887, opened 2026-04-07
feat: printing hook

oameye:
## Summary
Add a metadata-aware printing hook so downstream packages can customize `Base.show` output based on metadata attached to `BasicSymbolic` nodes.

This enables opt-in rendering for DSLs (e.g., sum prefixes or bracketed averages) without pirating.

## What changed
- Introduce `show_metadata(io, x)` and `show_metadata(io, x, ::Type{Ctx}, val)` hooks.
- Add `show_plain(io, x)` so hooks can reuse the default rendering without recursion.

## Example usage
```julia
struct SumIndices end

function SymbolicUtils.show_metadata(io::IO, x::SymbolicUtils.BasicSymbolic, ::Type{SumIndices}, val)
    print(io, "Σ(", val, ")")
    SymbolicUtils.show_plain(io, x)
    return true
end
```

## Example output
- `average(Σ(...))` can render as `Σ(i=1:N)⟨...⟩`
- `average(a)` can render as `⟨a⟩`

## Tests
- Added a test in `test/basics.jl` showing a custom metadata hook overriding output.

## Motivation
Downstream packages currently cannot alter printing based on metadata (only on the operator via `show_call`). This PR adds a general, non-pirating hook while keeping default behavior unchanged.

### DIFF
--- a/src/printing.jl
+++ b/src/printing.jl
@@ -1,4 +1,22 @@
-function Base.show(io::IO, x::BasicSymbolic{TreeReal})
+"""
+    show_metadata(io, x) -> Bool
+
+Hook for custom printing based on metadata. Return `true` if printing has been handled.
+Downstream packages can extend `show_metadata(io, x, ::Type{Ctx}, val)` for their
+metadata context and call `show_plain(io, x)` to bypass this hook.
+"""
+function show_metadata(io::IO, x::BasicSymbolic)
+    md = metadata(x)
+    md isa AbstractDict || return false
+    for (ctx, val) in md
+        show_metadata(io, x, ctx, val) && return true
+    end
+    return false
+end
+
+show_metadata(::IO, ::BasicSymbolic, ::DataType, @nospecialize(val)) = false
+
+function show_plain(io::IO, x::BasicSymbolic{TreeReal})
     @match x begin
         BSImpl.Sym(; name) => Base.show_unquoted(io, name)
         BSImpl.Const(; val) => begin
@@ -15,7 +33,7 @@ function Base.show(io::IO, x::BasicSymbolic{TreeReal})
     end
 end
 
-function Base.show(io::IO, x::BasicSymbolic)
+function show_plain(io::IO, x::BasicSymbolic)
     @match x begin
         BSImpl.Sym(; name) => Base.show_unquoted(io, name)
         BSImpl.Const(; val) => begin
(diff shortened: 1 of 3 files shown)


oameye:
@AayushSabharwal Can you have a look at this PR?

src/printing.jl
```diff
@@ -1,4 +1,22 @@
...
+"""
+    show_metadata(io, x) -> Bool
+
+Hook for custom printing based on metadata. Return `true` if printing has been handled.
+Downstream packages can extend `show_metadata(io, x, ::Type{Ctx}, val)` for their
+metadata context and call `show_plain(io, x)` to bypass this hook.
+"""
+function show_metadata(io::IO, x::BasicSymbolic)
+    md = metadata(x)
+    md isa AbstractDict || return false
```
AayushSabharwal:
```suggestion
    md isa Base.ImmutableDict{DataType, Any} || return false
```

Asserting concrete types tends to give the compiler more information

src/printing.jl
```diff
@@ -1,4 +1,22 @@
...
+    for (ctx, val) in md
+        show_metadata(io, x, ctx, val) && return true
+    end
+    return false
+end
+
+show_metadata(::IO, ::BasicSymbolic, ::DataType, @nospecialize(val)) = false
```
AayushSabharwal, line 17: In `despecialize.jl` could you define extra implementations of this method (at least 3). This prevents the compiler from specializing generated code on the number of methods, which avoids excessive invalidation when downstream packages implement this.

AayushSabharwal (commented):
While I am not a fan of variable metadata, I think this is the cleanest way to implement custom printing. I just have one question: I thought SQA implemented its own wrapper types over `BasicSymbolic`? Could you expand on why defining `Base.show` on them is not sufficient?

oameye:
Thanks a lot for having a look at this PR!

> I thought SQA implemented its own wrapper types over `BasicSymbolic`? Could you expand on why defining `Base.show` on them is not sufficient?

On the main branch (targeting SymbolicUtils v3), we had `AvgSym <: CNumber` and `const Average = BasicSymbolic{<:AvgSym}`. Since v3 encoded the symtype in `T` of `BasicSymbolic{T}`, `Average` was a dispatchable SQA-parameterized type and `Base.show` could be defined on it cleanly.
In v4, `T` changed meaning: it now represents the vartype (`SymReal`, `SafeReal`, etc.) rather than the symtype. `AvgSym` is not a valid vartype, so averages are now plain `BasicSymbolic{SymReal}`, indistinguishable at the type level from any other symbolic expression. There is nothing left to dispatch on, hence the need for a hook like `show_metadata`.

oameye:
Do we need docs?

oameye:
all integration tests seems to be unrelated

PR #891, opened 2026-04-14
refactor: [AI] limit `IRStructure` printing to 50 lines

AayushSabharwal:
It can get pretty insane if it always prints the full IR.

### DIFF
--- a/src/irstructure.jl
+++ b/src/irstructure.jl
@@ -146,7 +146,16 @@ function _print_ssa_var(io::IO, i::Int)
     printstyled(io, "%", i; color = :yellow)
 end
 
-function Base.show(io::IO, ir::IRStructure)
+"""
+    $TYPEDSIGNATURES
+
+Print the IR representation of `ir` to `io`, showing at most `limit` statements.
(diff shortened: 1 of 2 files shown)


src/irstructure.jl
```diff
@@ -212,16 +223,39 @@ function Base.show(io::IO, ir::IRStructure)
...
+            print(io, "  ⋮ ($remaining more statement$(remaining == 1 ? "" : "s") omitted; use `print_ir` to show all)")
+            println(io)
+            return
+        end
         print(io, "  ")
         _print_ssa_var(io, new_idx[i])
         print(io, " = ")
         print_expr(i)
         println(io)
+        printed += 1
```
DhairyaLGandhi, line 240: Could probably limit the statements with `min(limit, n)` and `n > limit && print(io, "remaining ...")` to be easier?

DhairyaLGandhi:
Thanks for the quick turnaround!

src/irstructure.jl
AayushSabharwal: Not every loop iteration prints a line. It prints a sort of "compressed" IR, where statements with only one use are inlined.

PR #895, opened 2026-04-19
Coerce FNTYPE_SEED to UInt to unbreak 32-bit Julia precompile

### LINKED ISSUE #894: Precompile fails on 32-bit Julia: `FNTYPE_SEED` is a raw UInt64 literal, truncates to UInt when stored in `hash::UInt`
## Bug

On 32-bit Julia (`UInt == UInt32`), loading SymbolicUtils fails at precompile time:

```julia-repl
julia> using SymbolicUtils
ERROR: LoadError: InexactError: trunc(UInt32, 0x8b4142913acd1c23)
 [6] hash_bsimpl(...)  @ SymbolicUtils hashconsing.jl:467
...
[22] top-level scope  @ SymbolicUtils/src/SymbolicUtils.jl:178  # @compile_workload
```

Fires inside the `@compile_workload` block, so every downstream package fails to precompile on 32-bit too.

## Root cause

`src/hashconsing.jl:310` declares `FNTYPE_SEED` as a raw `UInt64` literal, unlike `SYM_SALT`/`DIV_SALT` on lines 225–226 which correctly coerce with `% UInt`:

```julia
const FNTYPE_SEED = 0x8b414291138f6c45          # line 310 — bug
const SYM_SALT    = 0x4de7d7c66d41da43 % UInt   # line 225 — correct
```

On 32-bit, `FNTYPE_SEED` is `UInt64`. Its only use (line 351) is `hash(...)::UInt ⊻ FNTYPE_SEED` — the `::UInt` only constrains the inner `hash` call, so the XOR widens the result to `UInt64`, which then fails to fit back into a `::UInt` field.

Sanity check: `0x8b414291138f6c45 ⊻ 0x8b4142913acd1c23 = 0x0000000029427066` — upper 32 bits unchanged, which is exactly what XOR'ing a zero-extended `UInt32` into a `UInt64` looks like.

## Fix

```diff
-const FNTYPE_SEED = 0x8b414291138f6c45
+const FNTYPE_SEED = 0x8b414291138f6c45 % UInt
```

No-op on 64-bit; truncates the constant to `UInt32` on 32-bit. Happy to send a PR.

## Versions

SymbolicUtils 4.24.2 (also 4.21.0); Julia 1.12 stable and nightly; Linux x86 and Windows x86. Not reproduced on any x64 target. Hit initially in CI for [ReversePropagation.jl](https://github.com/JuliaIntervals/ReversePropagation.jl).

## CI note

This regression escaped because `ci.yml` tests only `ubuntu-latest × x64` — no `arch` key in the matrix, so 32-bit targets are never exercised. Adding `arch: [x64, x86]` (with a macOS-x86 exclude) would have caught it.

dpsanders:
## Summary

One-line fix for #894: `FNTYPE_SEED` (`src/hashconsing.jl:310`) was a bare `UInt64` literal. Its only use site (`hash(...)::UInt ⊻ FNTYPE_SEED` at line 349) widens the XOR result to `UInt64` on 32-bit Julia, which then fails to fit back into a `::UInt` (= `UInt32`) hash field — so `using SymbolicUtils` errors out with `InexactError: trunc(UInt32, 0x8b4142913acd1c23)` during `@compile_workload`.

```diff
-const FNTYPE_SEED = 0x8b414291138f6c45
+const FNTYPE_SEED = 0x8b414291138f6c45 % UInt
```

Matches the style already used for `SYM_SALT`/`DIV_SALT` (lines 225-226). No-op on 64-bit, truncates the constant to `UInt32` on 32-bit.

See #894 for the full diagnostic (XOR fingerprint, versions, downstream impact).

Closes #894.

## CI note (not in this PR)

This slipped through because `ci.yml` runs only on `ubuntu-latest × x64` (no `arch` key in the matrix). Adding `arch: [x64, x86]` (with a macOS-x86 exclude) would catch this class of 32-bit regressions. Not touching that here — happy to open a separate PR if the maintainers want it.

### DIFF
--- a/src/hashconsing.jl
+++ b/src/hashconsing.jl
@@ -307,7 +307,7 @@ implementation, which uses `objectid`, which changes across runs.
 """
 hash_addmulvariant(x::AddMulVariant.T, h::UInt) = hash(x === AddMulVariant.ADD ? 0x6d86258fc9cc0742 : 0x5e0a17a14cd8c815, h)
 
-const FNTYPE_SEED = 0x8b414291138f6c45
+const FNTYPE_SEED = 0x8b414291138f6c45 % UInt
 
 """
     $TYPEDSIGNATURES


ChrisRackauckas:
open a PR to add 32-bit as well, that would be helpful, or just add here

dpsanders:
You mean to add it to CI?

ChrisRackauckas:
yes

PR #900, opened 2026-04-28
feat: [AI] use new `OrderedDiGraph` for `IRStructure`

AayushSabharwal:
This allows `outneighbors` to correspond with `arguments`. Codegen can subsequently be refactored to only depend on the graph and `operation(ir.symbols[i])`.

### DIFF
--- a/src/ordered_digraph.jl
+++ b/src/ordered_digraph.jl
@@ -1,0 +1,69 @@
+"""
+    $TYPEDEF
+
+A directed graph that wraps `Graphs.SimpleDiGraph` and remembers edge insertion order.
+`Graphs.outneighbors(g, v)` returns the out-neighbors of `v` in the order edges were
+added, rather than in sorted order.
+
+All other graph operations (`inneighbors`, `topological_sort_by_dfs`, etc.) are
+delegated to the underlying `SimpleDiGraph`.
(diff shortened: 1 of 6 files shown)


src/ordered_digraph.jl
```diff
@@ -0,0 +1,69 @@
...
+    $TYPEDEF
+
+A directed graph that wraps `Graphs.SimpleDiGraph` and remembers edge insertion order.
+`Graphs.outneighbors(g, v)` returns the out-neighbors of `v` in the order edges were
+added, rather than in sorted order.
+
+All other graph operations (`inneighbors`, `topological_sort_by_dfs`, etc.) are
+delegated to the underlying `SimpleDiGraph`.
+"""
+struct OrderedDiGraph <: Graphs.AbstractGraph{Int}
```
oscardssmith, line 11: IMO SymbolicUtils is the wrong place for this (and I'm also not sure this is a good implementation)

AayushSabharwal: New package? Also, what's wrong with the implementation? One thing I was considering is using `OrderedSet{Int}` instead.

AayushSabharwal:
Opening a new PR with the new new graph.

PR #902, opened 2026-04-30
feat: [AI] use new graph for `IRStructure`

AayushSabharwal:
The new graph is built specifically for the use cases that `IRStructure` cares about, and with some behavior (ordered `outneighbors`) which existing graph types do not provide. This may later be moved to a separate package once we settle on the design and it is extensively tested and benchmarked downstream.

### DIFF
--- a/src/faster_code.jl
+++ b/src/faster_code.jl
@@ -46,7 +46,7 @@ mutable struct CodegenState{T}
     In such a case, if we assume that the index of `x + y` in `ir` is `4`, then
     `cache[4] == Symbol("%3")`.
     """
-    const cache::Dictionary{Int, Symbol}
+    const cache::Dictionary{Int32, Symbol}
     """
     Rewrite rules, similar to `NameState`.
     """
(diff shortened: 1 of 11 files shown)


src/faster_code.jl
```diff
@@ -267,7 +266,7 @@ function fast_toexpr(sym::CodegenPrimitive, ir::IRStructure{T}, rewrites::Dict{A
 end
 
 """
-    function codegen_function!(fn::F, cs::CodegenState{T}, expr::BasicSymbolic{T}, expr_idx::Int) where {F, T}
+    function codegen_function!(fn::F, cs::CodegenState{T}, expr::BasicSymbolic{T}, expr_idx::Int32) where {F, T}
```
oscardssmith:
```suggestion
    function codegen_function!(fn::F, cs::CodegenState{T}, expr::BasicSymbolic{T}, expr_idx::Integer) where {F, T}
```
no reason to over-specialize.

src/irstructure.jl
```diff
@@ -570,11 +537,11 @@ clear_cache!(sub::IRSubstituter) = empty!(sub.cache)
 Perform the substitution on element `idx` in the IR, returning the index of the new
 element.
 """
-function substitute_ir!(sub::IRSubstituter{Fold, T}, idx::Int) where {Fold, T}
+function substitute_ir!(sub::IRSubstituter{Fold, T}, idx::Int32) where {Fold, T}
     (; rules, filterer, ir) = sub
 
     # Check the cache, filter, and rules for `idx`
-    cached = get(sub.cache, idx, 0)
+    cached = get(sub.cache, idx, zero(Int32))
```
oscardssmith, line 658:
```suggestion
    cached = get(sub.cache, idx, zero(T))
```

AayushSabharwal: `T` here is from `BasicSymbolic{T}`, not the integer type.

AayushSabharwal:
Tests pass as much as they will given the current state of MTK. The MTK failures are unrelated to this PR. I'll rebase the fixups and merge the PR.

PR #906, opened 2026-05-01
Fix poly_to_gcd_form to produce concrete coefficient eltypes

ChrisRackauckas-Claude:
## Summary

`poly_to_gcd_form` re-packs a polynomial via broadcast — `Integer.(MP.coefficients(p))`, `rationalize.(...)`, `(complex ∘ float).(...)`, `float.(...)`. Broadcast preserves the abstractness of the input vector's eltype: `Integer.(::Vector{Integer})` returns `Vector{Integer}`, not `Vector{Int}`. So when `to_poly!` produces a `DP.Polynomial` with an abstract coefficient eltype (which it does in plenty of `simplify_div` paths), we feed `MP.gcd` polynomials whose type parameter is `Integer` / `Signed` / `AbstractFloat` / `Real`, and `MP.gcd` crashes:

- in `isolate_variable`, the reconstructed `Vector{Term{<:Any, M}}` hits a `MethodError(Term{T,M} where T, (::Vector{Term{<:Any, M}},))` when MP tries to dispatch `Polynomial{V,M}(::Vector{Term})` through the generic fallback at `DP.poly.jl:100` (`Polynomial{V,M}(α) = Polynomial(_Term{V,M}(α))`), which in turn calls `Term{T,M}(::Vector{Term})` — a method that doesn't exist;
- or, when both sides were `AbstractFloat` / `Signed`, internal arithmetic eventually falls through to `gcd(::Float64, ::Float64)` (intentionally unimplemented).

Re-pack each branch's coefficients into a `Vector{T}` where `T` is the `promote_type` over the concrete element types — or a sane default for the empty-input edge case. The resulting `DP.Polynomial` then always carries a concrete type parameter and `MP.gcd` runs cleanly.

## Surfaced via SymbolicIntegration.jl

`simplify(integral_diff; expand=true)` inside [SymbolicIntegration.jl's difficult-test verifier](https://github.com/JuliaSymbolics/SymbolicIntegration.jl/blob/main/test/rundifficulttests.jl#L107) crashes with the `MethodError(MP.Term{T,M} where T, (::Vector{Term{_A, M} where _A},))` for several rationals in the Apostol corpus, e.g.

- `(2 - x + 2x^2 - x^3 + x^4) / ((-1 + x)·(2 + x^2)^2)`
- `1/(-1 + x^2)^2`
- `x^2 / (2 + 2x + x^2)^2`

The integration itself succeeds; only the symbolic equivalence check throws. With this patch the same `simplify` returns a real expression instead.

## Test plan

- [x] Locally reduced to a minimal bad input — abstract `Polynomial{V,M,Signed}` × `Polynomial{V,M,AbstractFloat}` going through `MP.gcd` reproduces the failure on `master`.
- [x] After the fix, the same `MP.gcd` call runs to completion and `simplify((x^4 - x^3 + 2x^2 - x + 2)/((x-1)·(x^2+2)^2) - <expected antiderivative>; expand=true)` no longer throws.
- [ ] CI green here.

### DIFF
--- a/src/polyform.jl
+++ b/src/polyform.jl
@@ -220,15 +220,31 @@ function poly_to_gcd_form(p::PolynomialT)
         any_complex |= c isa Complex
         all_int || all_rat || break
     end
-    if all_int
-        return DP.Polynomial(Integer.(MP.coefficients(p)), MP.monomials(p))
+    cs = if all_int
+        Integer.(MP.coefficients(p))
     elseif all_rat
-        return DP.Polynomial(rationalize.(MP.coefficients(p)), MP.monomials(p))
+        rationalize.(MP.coefficients(p))
     elseif any_complex
-        return DP.Polynomial((complex ∘ float).(MP.coefficients(p)), MP.monomials(p))
+        (complex ∘ float).(MP.coefficients(p))
     else
-        return DP.Polynomial(float.(MP.coefficients(p)), MP.monomials(p))
+        float.(MP.coefficients(p))
     end
+    # Broadcast preserves the abstractness of the input vector's eltype:
+    # `Integer.(::Vector{Number})` returns `Vector{Number}` if the values
+    # are heterogeneous concrete subtypes of Integer (e.g. `Int8 + Int64`
+    # broadcasts to `Vector{Signed}`). The resulting `DP.Polynomial` then
+    # carries an abstract type parameter (`Integer`/`Signed`/`AbstractFloat`/
+    # `Real`), which crashes `MP.gcd`'s `isolate_variable` reconstruction.
+    # Narrow to a concrete eltype here when needed; on the homogeneous fast
+    # path (eltype already concrete) this is a single `isconcretetype` check
+    # and no extra allocation.
+    if !isconcretetype(eltype(cs))
+        T = isempty(cs) ? (all_int ? Int : all_rat ? Rational{Int} :
+                           any_complex ? ComplexF64 : Float64) :
+            mapreduce(typeof, promote_type, cs)
+        cs = Vector{T}(cs)
+    end
+    return DP.Polynomial(cs, MP.monomials(p))
 end
 
 function safe_gcd(p1::Union{PolyVarT, PolynomialT}, p2::Union{PolyVarT, PolynomialT})
--- a/test/polyform.jl
+++ b/test/polyform.jl
@@ -67,6 +67,109 @@ end
     @eqtest simplify_fractions((x * y + (1//2) * x) / (2 * x)) == (1//2 + y) / 2
 end
 
+import DynamicPolynomials as DP
+import MultivariatePolynomials as MP
+using SymbolicUtils: PolynomialT, poly_to_gcd_form, MonomialOrder
+
+# `to_poly!` returns `PolynomialT = DP.Polynomial{V,M,Number}` — the
+# coefficient eltype is the abstract `Number`. The previous broadcast form in
+# `poly_to_gcd_form` (`Integer.(coeffs)`, `rationalize.(...)`, `float.(...)`,
+# `(complex ∘ float).(...)`) preserved abstractness whenever the underlying
+# concrete element types were *heterogeneous* — e.g.
+# `Integer.(Number[Int8(1), Int64(-2), Int64(1)])` returns
+# `Vector{Signed}` rather than `Vector{Int64}`. The resulting
+# `DP.Polynomial` then carried an abstract type parameter
+# (`Signed` / `Integer` / `AbstractFloat` / `Real`), and `MP.gcd`'s
+# `isolate_variable` crashed downstream with a
+# `MethodError(Term{T,M} where T, (::Vector{Term{<:Any, M}},))`.
+#
+# Guard: each branch of `poly_to_gcd_form` must return a polynomial with a
+# *concrete* coefficient type even when the input vector is heterogeneous
+# within the relevant numeric kind.
+let v = only(DP.@polyvar __PolyToGcdFormTest__ monomial_order = MonomialOrder)
+    # CRITICAL: build the coefficient vector with the typed-literal form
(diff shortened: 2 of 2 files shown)


ChrisRackauckas-Claude:
Added a `test/polyform.jl` regression set:

1. **`poly_to_gcd_form gives concrete coefficient eltype`** — feeds `poly_to_gcd_form` `PolynomialT` (i.e. `DP.Polynomial{V,M,Number}`) inputs whose underlying concrete element types are heterogeneous within each numeric kind. The previous broadcast form `Integer.(::Vector{Number})` only collapses to `Vector{Int64}` when the input values are homogeneous; with `Number[Int8(1), Int64(-2), Int64(1)]` it falls back to `Vector{Signed}`, which is exactly the abstract-eltype path that crashed `MP.gcd`. Tests all four numeric branches (integer/rational/float/complex) plus a round-trip through `MP.gcd` that used to throw `MethodError(Term{T,M} where T, (::Vector{Term{<:Any, M}},))`.

2. **`simplify_fractions doesn't crash on heterogeneous coefficients`** — end-to-end smoke through the public API.

Verified by reverting `src/polyform.jl` to the previous broadcast form: the new testset reports **4 failures + 1 errored** (matching the `MethodError` we set out to fix). With the patch applied: **all 11 + 2 pass**.

`@testset` summary on patched code:
\`\`\`
poly_to_gcd_form gives concrete coefficient eltype                |   11     11  3.1s
simplify_fractions doesn't crash on heterogeneous coefficients    |    2      2  9.4s
\`\`\`

ChrisRackauckas-Claude:
## Performance / type-stability check

Reworked the fix as v2 (force-pushed) after the first version showed a real performance regression. The v2 keeps the original broadcast (fast path) and only narrows the eltype when broadcast came back abstract — so the homogeneous-coefficient hot path is byte-identical to the pre-fix code.

### Type stability (JET)

`poly_to_gcd_form`'s inferred return type is `Any` both before and after — the eltype is always determined at runtime (it depends on the actual coefficient values, not their declared types). `@report_call` reports nothing on either version. So this PR neither helps nor hurts top-level inferrability; what it changes is that the `output coefficient eltype` is now always `isconcretetype` instead of sometimes returning `Polynomial{V,M,Signed}` / `Polynomial{V,M,Rational}` / `Polynomial{V,M,AbstractFloat}` (which is what made `MP.gcd` blow up downstream).

| input shape | output coef eltype (pre-fix) | output coef eltype (this PR) |
|---|---|---|
| `Polynomial{V,M,Number}` of all-Int64 coefs | `Int64` ✓ | `Int64` ✓ |
| `Polynomial{V,M,Number}` of `Int8 + Int64` coefs | **`Signed`** ❌ | `Int64` ✓ |
| `Polynomial{V,M,Number}` of `Rational{Int8} + Rational{Int64}` coefs | **`Rational`** ❌ | `Rational{Int64}` ✓ |
| `Polynomial{V,M,Number}` of `Float32 + Float64` coefs | **`AbstractFloat`** ❌ | `Float64` ✓ |

### Microbenchmark — `poly_to_gcd_form`

| input | pre-fix (median, allocs, bytes) | this PR (v2) | Δ |
|---|---|---|---|
| homogeneous Int (3-term) | 2.89 μs / 12a / 448B | 2.89 μs / 12a / 448B | **0%** |
| heterogeneous Int (3-term) | 4.23 μs / 19a / 688B (broken) | 5.91 μs / 21a / 768B (correct) | +40% |
| homogeneous Rat (2-term) | 2.12 μs / 13a / 512B | 2.15 μs / 13a / 512B | **+1%** |
| heterogeneous Rat (3-term) | 5.31 μs / 24a / 848B (broken) | 6.46 μs / 27a / 992B (correct) | +22% |
| homogeneous Float (2-term) | 1.92 μs / 12a / 448B | 1.95 μs / 12a / 448B | **+1%** |
| heterogeneous Float (3-term) | 3.46 μs / 21a / 736B (broken) | 5.34 μs / 24a / 832B (correct) | +54% |
| **large degree-100 polynomial (homogeneous)** | 28.6 μs / 104a / 2.62KiB | 28.3 μs / 104a / 2.62KiB | **0%** |

Heterogeneous-coefficient inputs pay ~20–55% extra (the cost of the `Vector{T}(cs)` narrowing pass that's the actual fix); homogeneous inputs are within noise of the pre-fix code.

### Downstream — `simplify_fractions` (the public API path)

| input | pre-fix | this PR (v2) | Δ |
|---|---|---|---|
| `(x + y)*(x - y) / (x + y)` | 32.3 μs / 99a / 3.22KiB | 37.2 μs / 99a / 3.22KiB | +15% |
| rational coefs | 157.8 μs / 1053a / 36.91KiB | 159.4 μs / 1053a / 36.91KiB | +1% |
| float coefs | 160.0 μs / 1081a / 36.64KiB | 158.4 μs / 1081a / 36.64KiB | -1% |
| multivariate cancel | 526.9 μs / 5839a / 302.13KiB | 529.7 μs / 5839a / 302.13KiB | 0% |
| nested fractions | 50.5 μs / 164a / 5.22KiB | 47.8 μs / 164a / 5.22KiB | -5% |
| large degree numerator | 301.4 μs / 2191a / 149.90KiB | 295.8 μs / 2191a / 149.90KiB | -2% |
| `3*x^x` deep mul | 129.9 μs / 831a / 30.31KiB | 126.8 μs / 831a / 30.31KiB | -2% |

Allocation count and bytes are unchanged across all tested paths; wall-time delta sits in the ±5% noise band except for the trivial `(x+y)·(x-y)/(x+y)` case (+5 μs, ≈ the cost of the extra `isconcretetype` check on a single very fast call).

JET / benchmark scripts: [`/tmp/jet_check.jl`](https://github.com/JuliaSymbolics/SymbolicUtils.jl) + [`/tmp/bench_downstream.jl`](https://github.com/JuliaSymbolics/SymbolicUtils.jl) (run with `JET.@report_call` and `BenchmarkTools.@benchmark`).

AayushSabharwal (approved):

PR #911, opened 2026-05-06
 feat: make codegen work on non-canonical `IRStructure`

AayushSabharwal:
This allows codegen passes to use `replace_node!` to efficiently update the IR, and then codegen from the mutated IR.


DhairyaLGandhi:
We can't rely on `operation(ir[idx]) == operation(ir_new[idx])` since we might have transformed it into something else at that index, right?

AayushSabharwal:
Is `ir_new` a mutated copy of `ir`? If so, yes.

PR #922, opened 2026-05-18
expand: fold literal `complex(re, im)` Terms in to_poly!

### LINKED ISSUE #921: simplify(...; expand=true) doesn't reduce diffs with literal complex(re, im) Terms
Two arithmetically-identical diffs that should reduce to 0 under
`expand = true` don't:

```julia
using Symbolics
@variables Δ::Real z::Number

simplify(1im * Δ * z - z * Δ * 1im; expand = true)
# z*complex(0, Δ) + (0 - 1im)*z*Δ

simplify(2im * Δ * z - 2 * (1im * Δ * z); expand = true)
# z*complex(0, 2Δ) - 2z*complex(0, Δ)
```

In each case one arm lands in a literal `complex(0, Δ)` symbolic Term and
the other in a `(0 + 1im)` numeric coefficient. They never combine.

Reproduces in plain SymbolicUtils, no Symbolics needed:

```julia
using SymbolicUtils
@syms x::Real
c = Term{SymReal}(complex,
                  [Term{SymReal}(zero, [x]; type = Real), x];
                  type = Complex{Real})
simplify(c - 1im * x; expand = true)
# complex(zero(x), x) - 1im*x
```

As far as I can tell: `Symbolics.unwrap(::Complex{Num})`
(Symbolics/src/complex.jl, around line 21) produces a literal
`Term(complex, [re, im])` whenever the components are symbolic.
SymbolicUtils pattern-matches that form for the O(1)
`real(complex(re, im))` / `imag` / `conj` fast paths
(SymbolicUtils/src/methods.jl:484, 493, 503), but `to_poly!` in
`polyform.jl` has no case for `f === complex`, so under `expand` the
node ends up as an opaque `PolyVar` and the simplifier never sees the
imaginary part.

oameye:
Fixes #921.

`Term(complex, [re, im])` (from `Symbolics.unwrap(::Complex{Num})`) has no case in `to_poly!`, so `expand` treats it as an opaque `PolyVar`. Identically-zero diffs that contain one don't reduce:

```julia
simplify(1im * Δ * z - z * Δ * 1im; expand = true)
# was:  z*complex(0, Δ) + (0 - 1im)*z*Δ
# now:  0
```

Fix is a single case in `to_poly!`: treat `complex(re, im)` as `re + 1im * im` polynomially.

I originally tried this as a `simplify` rule but that broke downstream tests (Symbolics overloads, ODE solver) because the `Term(complex, ...)` literal is also the shape `Base.real/imag/conj` pattern-match for O(1) extraction and the shape `Complex{Num}` round-trips through. Keeping the rewrite local to `to_poly!` leaves that canonical form alone.

### DIFF
--- a/src/polyform.jl
+++ b/src/polyform.jl
@@ -120,6 +120,13 @@ function to_poly!(poly_to_bs::AbstractDict, bs_to_poly::AbstractDict, expr::Basi
                     MA.operate!(f, poly, to_poly!(poly_to_bs, bs_to_poly, arg))
                 end
                 return poly
+            elseif f === complex && length(args) == 2
+                # `Term(complex, [re, im])` (produced by
+                # `Symbolics.unwrap(::Complex{Num})` when either component is
+                # symbolic) is otherwise opaque to expansion. Treat it as
+                # `re + 1im * im` polynomially so identically-zero diffs that
+                # contain such a literal reduce under `simplify(...; expand=true)`.
+                return to_poly!(poly_to_bs, bs_to_poly, args[1] + 1im * args[2], recurse)
             else
                 if recurse
                     expr = BSImpl.Term{T}(f, map(expand, args); type, shape)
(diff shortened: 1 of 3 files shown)


AayushSabharwal (approved):

AayushSabharwal:
This seems to break a symbolic solver test. If I have time over the weekend, I'll try and take a look at it. The symbolic solvers are somewhat annoyingly capricious at times.

PR #924, opened 2026-05-18
fold conj, real, imag on the symbolic imaginary unit

oameye:
Three new `@match` arms in `Base.real`, `Base.conj`, `Base.imag` for `BasicSymbolic` that match `Sym(:im; type = Number)` structurally and fold to `0`, `-im`, `1`.

`Symbolics.IM` is `Sym{VartypeT}(:im; type = Number)`, used as a stand-in for `1im` so expressions stay inside `BasicSymbolic{<:Real}` algebra. Without these arms, `conj(IM)` / `real(IM)` / `imag(IM)` were opaque wrappers that `simplify` could not reduce, and downstream code (e.g. SQA's `qadjoint` / `inner_adjoint`) had to special-case `IM` themselves to avoid `conj(im)` leftovers.

Matching is purely structural (name + symtype), so SymbolicUtils picks up no Symbolics-side dependency. A same-named sym with a non-`Number` symtype is unaffected; the test covers that case.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -479,6 +479,8 @@ function Base.real(s::BasicSymbolic{T}) where {T}
     @match s begin
         BSImpl.Const(; val) => Const{T}(real(val))
         BSImpl.Term(; f, args) && if f === complex && length(args) == 2 end => args[1]
+        # Match `Symbolics.IM = Sym(:im; type = Number)` structurally.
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
Thanks for your attention to these edge cases. I think the better way to phrase this is to move the `const IM` here, and compare against that in the functions. Otherwise, SymbolicUtils is effectively relying on a convention introduced in Symbolics.jl, when the dependency should be the other way around.

PR #933, opened 2026-05-21
refactor: use new technique for stable type hashes across machines

AayushSabharwal:
Hashes of types are not stable across Julia versions, and more importantly across machines. This means there can be significant observable differences in how expressions print on different machines. It also leads to bugs that cannot be reproduced on another machine in case algorithms indirectly depend on the hash ordering of expressions. This function converts types to strings and encodes the hash in the generated function. Since the hash of a string is stable across machines, the same type should consistently get the same hash.

In benchmarks, the dynamic dispatch incurred by calling this function was found to be comparable to or even cheaper than `hash(::DataType)`, depending on the type hashed.

### DIFF
--- a/src/hashconsing.jl
+++ b/src/hashconsing.jl
@@ -321,67 +321,12 @@ end
 hash_shape(sh::Unknown, h::UInt) = hash(sh, h)
 
 """
-    $METHODLIST
-
-Custom hash functions for `vartype(x)`, since hashes of types defined in a module are not
-stable across machines or processes.
-"""
-vartype_hash(::Type{SymReal}, h::UInt) = hash(0x3fffc14710d3391a, h)
-vartype_hash(::Type{SafeReal}, h::UInt) = hash(0x0e8c1e3ac836f40d, h)
-vartype_hash(::Type{TreeReal}, h::UInt) = hash(0x44ec30357ff75155, h)
-
-"""
     $TYPEDSIGNATURES
 
 Custom hash functions for `AddMul.variant`, since it falls back to the `Base.Enum`
 implementation, which uses `objectid`, which changes across runs.
 """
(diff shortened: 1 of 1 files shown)


src/hashconsing.jl
```diff
@@ -430,6 +375,23 @@ function hash_metadata(m::MetadataT, h::UInt)
...
+It also leads to bugs that cannot be reproduced on another machine in case algorithms indirectly
+depend on the hash ordering of expressions. This function converts types to strings and encodes the
+hash in the generated function. Since the hash of a string is stable across machines, the same
+type should consistently get the same hash.
+
+In benchmarks, the dynamic dispatch incurred by calling this function was found to be comparable
+to or even cheaper than `hash(::DataType)`, depending on the type hashed.
+"""
+@generated function hash_type(::Type{T}, h::UInt) where {T}
+    return :(hash($(hash(string(T))), h))
```
KristofferC, line 392: I am not sure it is 100% allowed to call `string` like this in a generated function. Consider for example

```
julia> module M
       export Foo
       struct Foo end
       end
Main.M

julia> string(M.Foo)
"Main.M.Foo"

julia> using .M

julia> string(M.Foo)
"Foo"
```

The state of the system changes what `string` returns and that is typcally not good in a generator.

AayushSabharwal: Oh, yikes. Is there a way to get a "canonical" form? Also, does the result of `string` depend on the "perspective" of `Main` or the module where `string` is called or something else entirely?

AayushSabharwal: I think this is still better than the current approach, though? In the sense that executing the same code in the same order should get the same result? Once `string` is called, the result is baked into the function.

KristofferC: > Also, does the result of string depend on the "perspective" of Main or the module where string is called or something else entirely?

I think it is `Main`, ref https://github.com/JuliaLang/julia/pull/29466

AayushSabharwal:
Unfortunately benchmarks seem to be a bit unhappy

PR #937, opened 2026-05-26
make static arrays instead of `Array` for small arrays

KristofferC:
Code generated now looks like

```
__miscₛᵧₘ15 = (SymbolicUtils.Code.fill_sarr)((Val)((3, 3)), var"##cse#74", var"##cse#75", ...
...
__miscₛᵧₘ16 = (SymbolicUtils.Code.fill_sarr)((Val)((3,)), var"##cse#76", var"##cse#76", var"##cse#79")
var"##cse#80" = (\)(__miscₛᵧₘ15, __miscₛᵧₘ16)
```

avoids allocation in creating them and in linear solving:

```
# before
julia> @btime prob.f(similar(prob.u0), prob.u0, prob.p, 0.0);
  125.292 μs (407 allocations: 18.70 KiB)

# after
julia> z = similar(prob.u0);

julia> @btime prob.f(z, prob.u0, prob.p, 0.0);
  118.375 μs (69 allocations: 2.59 KiB)
  ```

### DIFF
--- a/src/faster_code.jl
+++ b/src/faster_code.jl
@@ -579,7 +579,7 @@ function codegen_function!(::Type{ArrayMaker{T}}, cs::CodegenState{T}, expr::Bas
         for ax in sh
             push!(sz_expr.args, length(ax))
         end
-        if all(<=(STATIC_ARRAY_LIMIT), sz_expr.args) && length(sz_expr.args) <= 2
+        if all(<=(STATIC_ARRAY_LIMIT), sz_expr.args) && length(sz_expr.args) <= 2 && _allocator === zeros
             result = Expr(:call, fill_sarr, Expr(:call, Val, sz_expr))
             for idx in SymbolicUtils.stable_eachindex(expr)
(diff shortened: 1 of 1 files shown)


AayushSabharwal (commented):
The idea itself looks great. However, ignoring `with_allocator` (removing/delaying the `codegen_allocator_call!`) is incorrect. If an allocator is present, some other part of codegen (e.g. `ArrayMaker`) expects the result to be written to it. Ignoring this will then cause incorrect results.

KristofferC:
So only do this when the allocator is `zeros` (the default)?

AayushSabharwal:
Yes

ChrisRackauckas:
Do we need to use StaticArrays or could we just use tuples and TupleLU?

AayushSabharwal:
Typically if someone generates a `\`, they expect to get an array out. TupleLU would return a tuple, which might break people's expectations.

PR #942, opened 2026-05-28
Avoid float in symbolic integer powers when not necessary

fgerick:
This fixes https://github.com/JuliaSymbolics/Symbolics.jl/issues/1875, which is nice for exact simplifications of larger expressions.

### DIFF
--- a/src/symbolic_ops/pow.jl
+++ b/src/symbolic_ops/pow.jl
@@ -96,10 +96,12 @@ function ^(a::BasicSymbolic{T}, b::Union{AbstractArray{<:Number}, Number, BasicS
             end
             BSImpl.Term(; f, args) && if f === sqrt && (safe_isinteger(b) && Int(b) % 2 == 0 || b isa Rational && numerator(b)%2 == 0) end => begin
                 exp = safe_isinteger(b) ? (Int(b) // 2) : (b::Rational // 2)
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
Thank you! Please add a test so we can know if this behavior regresses. The test should fail without this PR and pass with it.

fgerick:
Before it was just a fix for `sqrt` and `cbrt`, but I think it is nicer to have this work for other rational/integer powers as well, which is what the last commit includes + a test for these additions.

AayushSabharwal:
The test failures are concerning

fgerick:
How can I do the integration tests locally to see if things work as they should? I simply tested SymbolicUtils.jl.

AayushSabharwal:
What I do is create an environment with Symbolics and this PR. Then run Symbolics tests to figure out what's going on. I typically run the specific failing tests by pasting them into the REPL. I think the Symbolics failures are identical to the others, so just fixing those should be enough.

fgerick:
I changed how floats are dealt with and also changed how coefficients are treated. Locally it showed me that the Symbolics.jl tests passed, with one `@test_broken` in the groebner extension no longer being broken.

fgerick:
Okay the way the factors are dealt with is not really how it should be yet. I will have a look at that when I have time.

PR #943, opened 2026-05-28
Avoid force inlining of `getindex` which is a big function

KristofferC:
This makes functions that call `getindex` (especially multiple times) become very bloated. 
A prime example of this is `cross` which calls it 12 times. The change in compile time can be seen in:

```julia-repl
julia> @time precompile(Tuple{typeof(LinearAlgebra.cross), Symbolics.Arr{Symbolics.Num, 1}, SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymbolicUtils.SymReal}})
  0.676814 seconds (4.16 M allocations: 188.066 MiB, 15.94% gc time, 100.00% compilation time)
true

julia> @time precompile(Tuple{typeof(LinearAlgebra.cross), Symbolics.Arr{Symbolics.Num, 1}, SymbolicUtils.BasicSymbolicImpl.var"typeof(BasicSymbolicImpl)"{SymbolicUtils.SymReal}})
  0.134154 seconds (398.82 k allocations: 20.483 MiB, 99.99% compilation time)
true
```

There is really no point in thinking about bounds checking for these operations because they won't SIMD and the overhead is negligible.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -1650,6 +1650,9 @@ function promote_shape(::typeof(LinearAlgebra.cross), shx::ShapeT, shy::ShapeT)
     return ShapeVecT((1:3,))
 end
 
+function _cross_product(::Type{T}, x, y) where {T}
+    BSImpl.Const{T}(ArgsT{T}((x[2] * y[3] - x[3] * y[2], -x[1] * y[3] + x[3] * y[1], x[1] * y[2] - x[2] * y[1])))
+end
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
I did not know `@propagate_inbounds` force inlined. Honestly, I don't even know why I added the `@propagate_inbounds` 😅

PR #946, opened 2026-05-29
refactor: use `OrderedDict`/`OrderedSet` over `Dict`/`Set`

AayushSabharwal:
As part of an attempt to make things less dependent on hash-ordering


KristofferC:
We should rerun the benchmarks here after https://github.com/JuliaCollections/OrderedCollections.jl/pull/166 is merged and a new version is tagged.

PR #950, opened 2026-06-01
avoid hashing keys twice in accumulate operations on dicts

KristofferC:
In a benchmark like:

```julia
using BenchmarkTools, SymbolicUtils
using Random

begin
    @syms a b c d
    Random.seed!(123)
    funs = [*, /]
    atoms = [a, b, c, d, a^2, b^2, a^1.5, (b + c), b^c, 1, 2.0]
    function random_term(len; atoms, funs, fallback_atom=1)
            xs = rand(atoms, len)
            while length(xs) > 1
                xs = map(Iterators.partition(xs, 2)) do xy
                    x = xy[1]; y = get(xy, 2, fallback_atom)
                    rand(funs)(x, y)
                end
            end
            xs[]
        end
    exs = [random_term(5; atoms, funs) for _ in 1:50]
    @btime SymbolicUtils.mul_worker(SymReal, $exs);
end
```

I see

```
# Before
  16.831 μs (236 allocations: 6.22 KiB)
# After
  14.578 μs (222 allocations: 6.00 KiB)
``` 

Want to see what the full benchmark suite looks like. The savings of this of course depends on how expensive hashing the keys are.

### DIFF
--- a/src/symbolic_ops/addsub.jl
+++ b/src/symbolic_ops/addsub.jl
@@ -108,24 +108,24 @@ function (awb::AddWorkerBuffer{T})(terms) where {T}
                         AddMulVariant.ADD => begin
                             newcoeff = newcoeff .+ coeff
                             for (k, v) in dict
-                                result[k] = get(result, k, 0) + v
+                                _accumulate!(result, k, v)
                             end
                         end
                         AddMulVariant.MUL => begin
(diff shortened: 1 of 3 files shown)


AayushSabharwal:
This at least seems to be a measurable improvement

AayushSabharwal:
Superseded by #952

PR #951, opened 2026-06-02
feat: [AI] implement symbolic versions of some codegen primitives.

AayushSabharwal:
This adds symbolic expressions for limited versions of some codegen primitives (such as `Func`). This allows type-stable representations of those primitives, as well as their use in `IRStructure`.

### DIFF
--- a/src/symbolic_codegen_primitives.jl
+++ b/src/symbolic_codegen_primitives.jl
@@ -1,3 +1,36 @@
+function SymbolicUtils.promote_symtype(::Type{DestructuredArgs}, name_type::TypeT, elem_types::TypeT...)
+    return name_type
+end
+
+function SymbolicUtils.promote_shape(::Type{DestructuredArgs}, name_shape::ShapeT, @nospecialize(elem_shapes::ShapeT...))
+    @nospecialize name_shape
+    return name_shape
+end
+
+"""
+    symDestructuredArgs(name, elems)
+
+Construct a symbolic expression representing an array argument that is destructured into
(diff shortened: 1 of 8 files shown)


src/symbolic_codegen_primitives.jl
```diff
@@ -0,0 +1,121 @@
...
+    return rhs_shape
+end
+
+"""
+    symAssignment(lhs, rhs)
+
+Construct a symbolic expression representing `lhs = rhs`. Returns a
+`Term{T}(Code.Assignment, [lhs, rhs])` with `symtype` and `shape` equal to those of `rhs`.
+"""
+function symAssignment(lhs::BasicSymbolic{T}, rhs::BasicSymbolic{T}) where {T}
```
DhairyaLGandhi, line 70: Will this let me write assignments inside `IRStructure`?

AayushSabharwal: Yep. The idea is to use `symLet` and `symAssignment` combined, so the assignments in `symLet` are executed as side effects before the body.

oscardssmith:
Why does this add so much code and not let us delete any?

AayushSabharwal:
The old versions support a lot more stuff that we rely on.

oscardssmith:
how practical would it be to migrate all the old users over?

AayushSabharwal:
The simple fact that the old versions are structs which people `getproperty` into makes it borderline impossible imo

oscardssmith:
can we deprecate the structs at least?

AayushSabharwal:
These versions are not fully-fledged alternatives. They don't support everything that the structs do.

PR #952, opened 2026-06-03
Test OrderedDict wihtout double hash in accumulate perf

KristofferC:
On top of https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/946


AayushSabharwal:
This seems to perform well

PR #954, opened 2026-06-03
Improve performance of `query`

KristofferC:
For this benchmark

```julia
using SymbolicUtils: query, isdiv
@syms a b c d x y z
  ex = (x + y) * (x - z) + sin(x) * cos(y) + (x + y + z)^2 + x * y * z
@btime query(isdiv, $ex) 
```

we go from

```julia 
julia> @btime query(isdiv, $ex)   
  197.163 ns (14 allocations: 224 bytes)
``` 

to

```julia
julia> @btime query(isdiv, $ex) 
  56.989 ns (0 allocations: 0 bytes)
```


AayushSabharwal:
It's annoying that `do`-blocks end up being actively detrimental to performance at times.

KristofferC:
`do` stands for "`do` not use them" :P

PR #955, opened 2026-06-03
improve performance of `cse`

KristofferC:
avoid having to allocate a closure on every call

Best to view diff without whitespace

The benchmark + timings are:

```julia
  using SymbolicUtils.Code: cse
  @syms a b c d x y z
  ex = SymbolicUtils.expand((x + y + z)^10)
  @btime cse($ex);           

# Before
#   137.283 μs (2463 allocations: 95.03 KiB)

# After
#   106.120 μs (2117 allocations: 78.81 KiB)
```


AayushSabharwal:
Might as well, though we don't actually use this code path anymore 😅

KristofferC:
-.-'

PR #960, opened 2026-06-04
trivial performance tweaks to avoid redundantly computing things etc

KristofferC:
I don't have benchmarks for this but they are trivially doing less work and removes some "noise" from the code.


KristofferC:
```
codegen/deep_poly/deg=10:fast_toexpr	1.25 k allocs: 0.0553 MB	3.12 k allocs: 0.107 MB	0.518
```

That doesn't look great...

AayushSabharwal:
What was the fix?

AayushSabharwal:
This has a merge conflict

AayushSabharwal:
The merge conflict returns 😅

PR #965, opened 2026-06-09
Add `ifelse_eager` and `ifelse_branching` conditional operators

ChrisRackauckas-Claude:

## Summary

Adds two conditional operators next to `ifelse`, requested for the 3D-multibody / spherical-joint work in [Symbolics.jl#1887](https://github.com/JuliaSymbolics/Symbolics.jl/pull/1887). They build the *same* kind of symbolic expression as `ifelse` and only differ in **how code generation lowers them**:

- **`ifelse_eager(cond, x, y)`** — always lowers to an eager `ifelse` call: both branches are evaluated and `cond` selects between the computed values. It needs **no custom codegen** — its definition is `ifelse(cond, x, y)`, so the generic lowering already emits a plain eager call.
- **`ifelse_branching(cond, x, y)`** — always lowers to an `if`/`else`, so the **untaken branch is never evaluated, even under CSE**. Useful when a branch is only valid when its condition holds (divides by a quantity that is zero otherwise, indexes into something that doesn't exist, errors, …).

`ifelse` stays the default; the intent is that it eventually picks a lowering strategy via a cost heuristic while these two pin a specific strategy.

## Implementation

- `methods.jl`: the two functions, their `BasicSymbolic` term construction, and `promote_symtype`/`promote_shape` (delegating to `ifelse`'s — condition must be `Bool`, branches share a shape).
- `faster_code.jl` / `code.jl`: a `codegen_function!` (fast path) and `function_to_expr` (slow path) for `ifelse_branching` that emit each branch into its own codegen scope (`enter_scope`/`exit_scope!`) so it lands inside the arm; plus `cse_inside_expr(_, ::typeof(ifelse_branching)) = false` so CSE doesn't hoist the branches out. `ifelse_eager` needs none of this — it lowers via the generic fallback.
- `simplify_rules.jl`: the same literal-condition / equal-branch folds `ifelse` has.
- Exported from `SymbolicUtils`.

## Tests

New `test/conditionals.jl` (wired into `runtests.jl`): construction/symtype/shape parity with `ifelse`, lowering forms (`:if` vs `:call`), `simplify` folding, and behavioral checks that `ifelse_branching` never runs the untaken branch (compiled via `cse` + `RuntimeGeneratedFunction`) while `ifelse_eager` does. All pass locally.

The `Num`/`Arr` wrappers, differentiation and linearity/sparsity integration live in the companion Symbolics.jl PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
 authors = ["Shashi Gowda"]
-version = "4.34.3"
+version = "4.35.0"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/src/SymbolicUtils.jl
+++ b/src/SymbolicUtils.jl
@@ -132,6 +132,7 @@ export @makearray
 include("arraymaker.jl")
 
 # Methods on symbolic objects
+export ifelse_eager, ifelse_branching
 PrecompileTools.@recompile_invalidations begin
 using SpecialFunctions, NaNMath
 include("methods.jl")
--- a/src/code.jl
+++ b/src/code.jl
@@ -19,7 +19,8 @@ import SymbolicUtils: @matchable, BasicSymbolic, Sym, Term, iscall, operation, a
                       search_variables!, _is_index_variable, RangesT, IDXS_SYM, is_array_shape,
                       symtype, vartype, add_worker, search_variables!, @union_split_smallvec,
                       ArrayMaker, TypeT, ShapeT, SymReal, SafeReal, TreeReal, _unreachable, unwrap,
-                      AddMulVariant, _isone, _iszero, Fill, IRStructure, populate_ir!, FnType
+                      AddMulVariant, _isone, _iszero, Fill, IRStructure, populate_ir!, FnType,
+                      ifelse_eager, ifelse_branching
 using Moshi.Match: @match
 import SymbolicIndexingInterface: symbolic_type, NotSymbolic
 import Graphs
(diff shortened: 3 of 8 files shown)


ChrisRackauckas-Claude:
CI: SymbolicUtils own tests all green — `test (1, Core)`, `test (1, QA)`, `test (1.11, Core)`, `test (min, Core)`, Spell Check. Downstream `Symbolics.jl/Core`, `Symbolics.jl/Downstream`, `Symbolics.jl/D3TreesExt`, `Symbolics.jl/LuxExt` also pass (this change is additive — new exported operators + methods + `simplify` rules — so it is backward-compatible).

The one red downstream check, `Symbolics.jl/GroebnerExt`, is a **pre-existing failure on Symbolics master** (stale `@test_broken` markers that now pass after the `Polyform`-removal refactor) — it fails against the released SymbolicUtils too, unrelated to this PR. Fixed separately in JuliaSymbolics/Symbolics.jl#1889.

The companion Symbolics.jl PR (#1887) adds the `Num`/`Arr` wrappers, differentiation and linearity/sparsity rules, and re-exports these operators.

ChrisRackauckas-Claude:
Final CI: **27 pass, 2 fail**. Both failures are pre-existing downstream issues, not caused by this additive change:
- `Symbolics.jl/GroebnerExt` — stale `@test_broken` on Symbolics master (fixed in JuliaSymbolics/Symbolics.jl#1889).
- `ModelOrderReduction.jl/All` — fails on this PR *and* on the disjoint Symbolics-side PR (#1887), and the `IntegrationTest` workflow is red on Symbolics master; it is a downstream-package breakage independent of these conditional operators.

All SymbolicUtils own tests (`test (1/1.11/min, Core)`, `QA`) and the core Symbolics downstream (`Symbolics.jl/Core`, `Downstream`, `D3TreesExt`, `LuxExt`) are green.

ChrisRackauckas-Claude:
Added the CSE follow-up in `b566ecf`: **multiply-referenced `ifelse_branching` now shares a single `if`/`else`** instead of duplicating it per use site.

The root cause was that `cse_inside_expr = false` conflated "do not descend into the arguments" with "do not bind this node". The commit adds a second hook, `Code.cse_bind_expr(sym, f)`, consulted only when `cse_inside_expr` is `false`: when it returns `true`, CSE binds the node **as-is** (arguments un-CSEd) to a temporary, so all references resolve to one shared computation. `ifelse_branching` opts in; the default `false` keeps the existing fully-inline behaviour for `Operator`, `getindex`, and any other opt-outs (their `cse(ex) == ex` identity tests are unchanged and pass).

Generated code for `sin(z) + cos(z)` with `z = ifelse_branching(x > 0, f(x)^2, g(x))` is now:
```julia
var"##cse#1" = if 0 < x
        f(x) ^ 2
    else
        g(x)
    end
var"##cse#2" = sin(var"##cse#1")
var"##cse#3" = cos(var"##cse#1")
```
— verified by side-effect counting that the taken branch runs **once** (was twice) on both the scalar and array codegen paths, the untaken branch still never runs, and one-sided nesting is now **linear**, matching `ifelse` (previously exponential). One semantic boundary, noted in the tests: a conditional referenced inside *both* branches of an enclosing `ifelse_branching` is still emitted per branch — hoisting it out would evaluate it eagerly, which laziness forbids.

Ran locally: full `Pkg.test()` — `17771 pass, 3 broken (pre-existing), 0 fail`.

ChrisRackauckas-Claude:
Note: the CSE node-sharing follow-up mentioned above (commit `b566ecf`) never attached to this PR — GitHub dropped the push's synchronize event, so it was not part of the merge. Re-landed on current master as #966.

PR #963, opened 2026-06-05
add some more specialized hashing functions

KristofferC:
For a benchmark:

```julia
using BenchmarkTools, SymbolicUtils
using SymbolicUtils: BasicSymbolic, ArgsT, SymReal, hash_argsvec

@syms a b c d e

mk(xs...) = ArgsT{SymReal}(collect(BasicSymbolic{SymReal}, xs))
args3 = mk(a, b, c)          # Backing-backed (len <= 3)
args5 = mk(a, b, c, d, e)    # Vector-backed  (len > 3)

# warm element hash caches
for v in (args3, args5), el in v
    hash(el, UInt(0))
end

for (name, args) in (("arity3 (Backing)", args3), ("arity5 (Vector)", args5))
    println("\n== ", name, " ==")
    print("  BEFORE  Base.hash : "); @btime hash($args, UInt(0))
    print("  AFTER   loop      : "); @btime hash_argsvec($args, UInt(0), true)
end
```

with results:

```
== arity3 (Backing) ==
  BEFORE  Base.hash :   30.477 ns (0 allocations: 0 bytes)
  AFTER   loop      :   14.123 ns (0 allocations: 0 bytes)

== arity5 (Vector) ==
  BEFORE  Base.hash :   46.678 ns (0 allocations: 0 bytes)
  AFTER   loop      :   20.054 ns (0 allocations: 0 bytes)
```

PR #966, opened 2026-06-12
Bind `ifelse_branching` to a CSE temporary while keeping branches lazy

ChrisRackauckas-Claude:

## Summary

Follow-up to #965 (this was developed on that PR's branch but GitHub dropped the push's synchronize event, so it never attached before the merge — re-landed here on current master).

**Multiply-referenced `ifelse_branching` now shares a single `if`/`else`** instead of re-emitting it per use site. Previously, `cse_inside_expr = false` conflated "do not descend into the arguments" with "do not bind this node": keeping the branches lazy also un-bound the conditional, so e.g. `sin(z) + cos(z)` with `z = ifelse_branching(...)` evaluated the taken branch twice.

## Implementation

A second hook, `Code.cse_bind_expr(sym, f)` (default `false`), consulted only when `cse_inside_expr` is `false`. When `true`, `_cse_compute` binds the node **as-is** — arguments un-CSEd — to a CSE temporary; `cse!`'s id-cache makes every reference in the scope resolve to that one temporary. `ifelse_branching` opts in. `Operator`, `getindex` and any other opt-outs keep their exact prior fully-inline behaviour (their `cse(ex) == ex` identity tests are untouched and pass).

Generated code for the repro is now:
```julia
var"##cse#1" = if 0 < x
        f(x) ^ 2     # taken branch: runs once
    else
        g(x)         # untaken: never runs
    end
var"##cse#2" = sin(var"##cse#1")
var"##cse#3" = cos(var"##cse#1")
```

One semantic boundary (asserted with a comment in the tests): a conditional referenced inside *both* branches of an enclosing `ifelse_branching` is still emitted per branch — hoisting it above the conditional would evaluate it eagerly, which the laziness contract forbids. One-sided nesting is linear, matching `ifelse` (previously exponential in depth).

## Verification

- New testsets in `test/conditionals.jl`: side-effect counting shows the taken branch fires **once** (was twice) and the untaken branch never; the CSE structure contains exactly one bound `ifelse_branching`; one-sided nesting emits the same number of `if`s as `ifelse`; values match `ifelse` throughout.
- Full local `Pkg.test()` on this branch (on top of the #963 hashing changes): **17771 pass, 3 broken (pre-existing), 0 fail**.
- `cse_bind_expr` documented and added to the codegen manual alongside `cse_inside_expr`, plus docs entries for the two operators.

Version bumped to 4.36.0. Companion test on the Symbolics side: JuliaSymbolics/Symbolics.jl#1892 (blocked on this releasing).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
 authors = ["Shashi Gowda"]
-version = "4.35.0"
+version = "4.36.0"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/docs/src/manual/codegen.md
+++ b/docs/src/manual/codegen.md
@@ -49,4 +49,18 @@ called within the expression are pure. SymbolicUtils can and will change the num
 ```@docs
 SymbolicUtils.Code.cse
 SymbolicUtils.Code.cse_inside_expr
+SymbolicUtils.Code.cse_bind_expr
+```
+
+### Conditionals
+
+`ifelse` lowers to an `if`/`else` whose branches are still subject to CSE. Two variants pin
+the evaluation strategy: `ifelse_eager` always evaluates both branches, while
+`ifelse_branching` guarantees the untaken branch is never evaluated — its branch interiors
+are excluded from CSE (the conditional itself is still bound, see
+[`SymbolicUtils.Code.cse_bind_expr`](@ref)).
+
+```@docs
+SymbolicUtils.ifelse_eager
+SymbolicUtils.ifelse_branching
 ```
--- a/src/code.jl
+++ b/src/code.jl
@@ -1553,6 +1553,22 @@ end
 cse_inside_expr(sym, ::typeof(ifelse_branching)) = false
 
 """
+    $(TYPEDSIGNATURES)
+
+Only consulted when [`cse_inside_expr`](@ref) returns `false` for `sym`, which has operation
+`f`. Return `true` if CSE should still bind `sym` itself to a temporary variable — so that
+multiple references to it share a single computation — while leaving its arguments un-CSEd.
+The default `false` leaves the expression fully inline at each reference site.
+"""
+function cse_bind_expr(sym, f)
+    return false
+end
+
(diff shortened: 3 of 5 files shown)


ChrisRackauckas-Claude:
CI triage — **26/29 green** (all SymbolicUtils own-test matrices, `Symbolics.jl/Core`, and `Symbolics.jl/GroebnerExt` now that JuliaSymbolics/Symbolics.jl#1889 is merged). The 3 failures, each investigated and **none caused by this PR**:

1. **`ModelingToolkit.jl/InterfaceI`** — 1 failure of 1386: `test/structural_transformation/tearing.jl:111` gets a permuted (same-nnz) incidence matrix. Bisected locally with a minimal repro of exactly that assertion against three SymbolicUtils commits:
   | SymbolicUtils commit | contains | result |
   |---|---|---|
   | `abcbea0c` (#965 merge = released v4.35.0) | operators, no hashing change | ✅ pass |
   | `79d748b7` (master = + #963 hashing) | **no commits from this PR** | ❌ fail, identical pattern |
   | `22d0a15b` (this PR head) | + `cse_bind_expr` | ❌ fail, identical pattern |

   ⇒ caused by **#963 (kc/hashing)** changing `Term` hashes → hash-order-dependent variable ordering in MTK's tearing. **Note: v4.35.0 was tagged *before* #963 merged, so the breakage is latent on master and will surface downstream with the next SymbolicUtils release** (including the 4.36.0 proposed here). Filed on the MTK side with the repro (issue link to follow).

2. **`Symbolics.jl/Downstream`** — JET fails to *precompile* on Julia 1.12.6 (`MethodError: add_active_gotos!` in `Compiler` internals); the job's DI Test dies loading JET. A JET↔Julia-patch incompatibility, unrelated to SymbolicUtils content.

3. **`ModelOrderReduction.jl/All`** — long-standing pre-existing failure (red on #965, Symbolics#1887, and master IntegrationTest alike).

ChrisRackauckas-Claude:
The `ModelingToolkit.jl/InterfaceI` failure now has a ready fix: SciML/ModelingToolkit.jl#4620 (makes the tearing matrix assertion accept both valid maximal matchings, mirroring the `||` pattern the same file already uses; mechanism + bisect in SciML/ModelingToolkit.jl#4619). Once that merges, this downstream check goes green for any SymbolicUtils release that includes #963.

AayushSabharwal (approved):

PR #968, opened 2026-06-13
fix that `combine_fold` can be called with a `op=Fill` with non-symbolic args

KristofferC:
Trying to evaluate `op` with a non-symbolic used to crash. Now it falls back to `maketerm` instead.

I am not sure we should actually try to execute the `Fill` here though (via the non-symbolic `fill` I guess)...


AayushSabharwal:
Yeah this is a bit of an annoying case. `fold = Val(true)` is meant to "evaluate" the expression, so I _guess_ `Fill` should call `fill`. I avoided adding a `::Fill` method for this case since it's very easy to accidentally call it expecting an expression and getting a massive array.

PR #972, opened 2026-06-14
Don't error on `x^1` for non-matrix array bases

baggepinnen:
`^(a::BasicSymbolic, b)` runs the (array) base through `promote_shape(^, …)`, which rejects non-matrix arrays — but it does so even for the **identity exponent `b == 1`**. A non-matrix array can legitimately appear as a multiplicity-1 factor of a `Mul`, which `arguments` materializes as `base^1`; that then errors (e.g. during code generation) with:

```
ArgumentError: Matrices are the only arrays that can be raised to a power. Found array of shape 1:3.
```

Since `x^1 == x` for any shape, this returns the base unchanged up front, before the array-shape check. Surfaced via ModelingToolkit/Symbolics RHS codegen of a multibody model, where a 3-vector factor appeared as `vec^1`.

Adds a regression test (`test/methods.jl`).

cc @AayushSabharwal

### DIFF
--- a/src/symbolic_ops/pow.jl
+++ b/src/symbolic_ops/pow.jl
@@ -54,8 +54,14 @@ function ^(a::BasicSymbolic{T}, b::Union{AbstractArray{<:Number}, Number, BasicS
     if !_numeric_or_arrnumeric_symtype(a) || !_numeric_or_arrnumeric_symtype(b)
         throw(MethodError(^, (a, b)))
     end
-    isconst(a) && return Const{T}(^(unwrap_const(a), b))
     b = unwrap_const(unwrap(b))
+    # `x^1 == x` for any shape. The array path below goes through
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
I think that this is likely masking the issue fixed in #969. `rand(3) ^ 1` also errors so I'm not sure that this change is valid.

baggepinnen:
Superseded by #969 (merged), which fixes the same `vector ^ 1` codegen error at the root: the `Mul` constructor no longer builds `k ^ 1` for an array base (it reconstructs the summand without raising the vector to a power), so the identity power that `^` rejects for non-matrix arrays never gets constructed in the first place. That's a cleaner fix than guarding `^(x, 1)` in `pow.jl` as this PR did. Closing in favor of #969.

PR #978, opened 2026-06-19
add a fast path for scalarizing `Fill`

KristofferC:
This pattern is not so uncommon, it happens for example in `collect(A)` in `unhack_system` in some cases

For a benchmark like:

```julia
using SymbolicUtils
using SymbolicUtils: Fill, ShapeVecT, scalarize, _default_scalarize
using BenchmarkTools

@syms x::Real

N = 2000
f = Fill(ShapeVecT((1:N,)))
A = f(x)                       # ArrayOp whose value is `x` everywhere

print("scalarize (fast _scalarize_fill): "); @btime scalarize($A);
print("_default_scalarize (per-element): "); @btime _default_scalarize($f, $A, Val{false}());
```

this gives

```julia
scalarize (fast _scalarize_fill):   766.490 ns (3 allocations: 16.06 KiB)
_default_scalarize (per-element):   657.042 μs (23500 allocations: 1.81 MiB)
```

### DIFF
--- a/src/substitute.jl
+++ b/src/substitute.jl
@@ -672,6 +672,22 @@ function _default_scalarize(f, x::BasicSymbolic{T}, ::Val{toplevel}) where {T, t
     end
 end
 
+scalarization_function(::Fill) = _scalarize_fill
+# Fast path for scalarizing a constant `Fill` array
+function _scalarize_fill(f, x::BasicSymbolic{T}, v::Val{toplevel}) where {T, toplevel}
+    @nospecialize f
+    sh = shape(x)
+    is_array_shape(sh) || return _default_scalarize(f, x, v)
+    val = @match x begin
+        BSImpl.ArrayOp(; expr) => expr
+        _ => nothing
+    end
+    if val isa BasicSymbolic{T} && !is_array_shape(shape(val))
+        return fill(val, prod(length, sh))
+    end
+    return _default_scalarize(f, x, v)
+end
+
 scalarization_function(::Type{ArrayOp{T}}) where {T} = _scalarize_arrayop
 
 function _scalarize_arrayop(_, x::BasicSymbolic{T}, ::Val{toplevel}) where {T, toplevel}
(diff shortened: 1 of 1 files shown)


src/substitute.jl
```diff
@@ -672,6 +672,22 @@ function _default_scalarize(f, x::BasicSymbolic{T}, ::Val{toplevel}) where {T, t
...
+    @nospecialize f
+    sh = shape(x)
+    is_array_shape(sh) || return _default_scalarize(f, x, v)
+    val = @match x begin
+        BSImpl.ArrayOp(; expr) => expr
+        _ => nothing
+    end
+    if val isa BasicSymbolic{T} && !is_array_shape(shape(val))
+        return fill(val, prod(length, sh))
+    end
```
AayushSabharwal, line 687: Doesn't this drop the dimensionality? E.g. if `sh == [1:3, 1:3]`, I'd expect this to return a 3x3 matrix but `prod(length, [1:3, 1:3]) == 9` which means it returns a 9-vector.

KristofferC: But `default_scalarize` does this: https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/74be77251706005c53e104beed2b70d38ab861c1/src/substitute.jl#L665

AayushSabharwal: Oh wow, that looks like a bug.

KristofferC: 😄

KristofferC: When trying to MWE this it seems the `default_scalarize` code is actually correct and this new code indeed just had a bug:

```
julia> idxs = CartesianIndices((1:3, 1:3))
CartesianIndices((1:3, 1:3))

julia> mat = rand(3,3)
3×3 Matrix{Float64}:
 0.116252  0.174089  0.850512
 0.692029  0.354339  0.564779
 0.610849  0.516795  0.204187

julia> [mat[idx] for idx in idxs]
3×3 Matrix{Float64}:
 0.116252  0.174089  0.850512
 0.692029  0.354339  0.564779
 0.610849  0.516795  0.204187
```

I didn't know the shape of a comprehension worked like that.

Will fix, well spotted

AayushSabharwal: Oh, I completely forgot that we return `CartesianIndices`

PR #990, opened 2026-06-26
fix: make expression ordering less dependent on hash

AayushSabharwal:
- Fix `_sorted_args3` (TreeReal) mistakenly calling `_sorted_args2` (SafeReal), which threw a MethodError when `<ₑ` called `sorted_arguments` on TreeReal expressions
- Split ADD and MUL cases in `__sorted_args`: ADD now sorts arguments directly with `<ₑ` (a total order), replacing the previous `sortperm(degrees; lt=monomial_lt)` whose stable-sort tie-breaking was hash-dependent; MUL keeps its existing `sortperm(degrees)` which is already hash-independent
- Replace `hash(expr)`-based placeholder in `__default_degrees!` with a session-stable fallback: callable Sym operations use their name, named functions/types use `nameof`, and unknown callables use a fixed sentinel symbol
- Use `sorted_arguments` instead of `arguments` in the `<ₑ` structural fallback so that equal-degree expressions compare in canonical rather than hash-dependent order

### DIFF
--- a/src/ordering.jl
+++ b/src/ordering.jl
@@ -54,7 +54,22 @@ end
 end
 
 function __default_degrees!(degrees::DegreesT, expr::BasicSymbolic{T}) where {T}
-    push!(degrees, SmallV{Symbol}((Symbol(:zzzzzzz, hash(expr)),)) => 1.0)
+    @match expr begin
+        BSImpl.Term(; f) => begin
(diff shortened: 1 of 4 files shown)


AayushSabharwal:
MTK/Initialization failure is just due to a particularly bad choice of states in a severely overdetermined initialization.

PR #995, opened 2026-06-30
'`handle_let_pair!`: `sizehint!` dicts that rewrites gets assigned into

KristofferC:
this avoids somewhat frequest rehasing of these dicts


KristofferC:
Hm, thinking about it again, I'm a little bit worried this will explode the size of these dicts since they seem to always get bigger but are never allowed to shrink. It depends on how long these dicts are persisting...

AayushSabharwal:
They persist for however long it takes to generate the function. Also, looks like 1.10 doesn't support the `shrink` kwarg.

PR #1000, opened 2026-07-02
`handle_let_pair!`: memoize destructuring rewrites

KristofferC:
The same work was done twice here for the oop/iip codegen and the cache can also be used between different generated functions

This makes the creation time of `ODEProblem` for a fairly large model go from

```
3.473098 seconds (54.42 M allocations: 2.252 GiB)
```

```
2.860457 seconds (43.66 M allocations: 1.834 GiB)
```

Note that the cache introduced here was emptied before making any timings to avoid caching from a previous run influencing the timing result.

### DIFF
--- a/src/code.jl
+++ b/src/code.jl
@@ -22,6 +22,7 @@ import SymbolicUtils: @matchable, BasicSymbolic, Sym, Term, iscall, operation, a
                       AddMulVariant, _isone, _iszero, Fill, IRStructure, populate_ir!, FnType,
                       ifelse_eager, ifelse_branching
 using Moshi.Match: @match
+import TaskLocalValues: TaskLocalValue
 import SymbolicIndexingInterface: symbolic_type, NotSymbolic
 import Graphs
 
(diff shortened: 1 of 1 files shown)


KristofferC:
If having a global cache like this is uncomfortable (but I guess SU already have quite a few of those already), a cache could also be passed in from higher up and then MTK creates it and carries it around.

src/code.jl
```diff
@@ -872,16 +876,78 @@ end
...
+    # so mark the dict and skip the (identical) re-population on the second call.
+    marker = (DESTRUCTURING_MARKER, key)
+    len_before = length(st.rewrites)
+    get!(st.rewrites, marker, true)
+    length(st.rewrites) == len_before && return nothing
+
+    cache = DESTRUCTURING_REWRITES_CACHE[]
+    entry = get(cache, key, nothing)
+    if entry === nothing
+        exprs = Vector{Any}(undef, n)
```
AayushSabharwal: Why not `Vector{Expr}`?

src/code.jl
```diff
@@ -872,16 +876,78 @@ end
...
+        for i in 1:n
+            exprs[i] = destructured_elem_expr(name, d.inds[i])
+        end
+        # crude bound to avoid unbounded growth over a long session
+        length(cache) > 10_000 && empty!(cache)
+        cache[key] = DestructuringRewrites(elems, d.inds, exprs)
+    else
+        exprs = entry.exprs
+    end
+    sizehint!(st.rewrites, length(st.rewrites) + n; shrink = false)
```
AayushSabharwal: `shrink` doesn't exist on 1.10

KristofferC:
Updated

PR #1002, opened 2026-07-03
Harden Const hashconsing against payloads with non-Bool isequal

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.**

## Problem

`isequal_somescalar`'s fallback used `isequal(a, b)::Bool`. That throws for `Const` payloads whose comparison operators are *symbolic* and do not return a `Bool` — e.g. `CasADi.MX`, whose `==`/`isequal` return an `MX`.

This surfaces in ModelingToolkit.jl's CasADi dynamic-optimization backend (SciML/ModelingToolkit.jl#4706). When a symbolic parameter/variable is substituted with a raw `CasADi.MX`, `substitute` wraps the value in `Const{T}(value)` (both the leaf case in `substitute.jl` and folded results in `combine_fold`), which goes through hashconsing. When the new `Const(::MX)` lands in the same `WeakCacheSet` bucket as a cached numeric `Const`, hashconsing compares payloads via `isequal_somescalar(::MX, ::Float64)`, and the `::Bool` fallback throws:

```
TypeError: non-boolean (CasADi.MX) used in boolean context
```

Because it depends on a hash-bucket collision, it looks version-/run-dependent.

### Minimal reproduction (with a mock of `MX`'s contract)

```julia
struct MockMX <: Number; v::Float64; end
Base.isequal(::MockMX, ::Number) = MockMX(0.0)   # non-Bool, like CasADi.MX
# before this PR:
SymbolicUtils.isequal_somescalar(MockMX(2.0), 1.0)  # TypeError: non-boolean (MockMX) used in boolean context
```

## Fix

Route the fallback through `isequal_somescalar_fallback`, which:

- returns `false` on a payload **type mismatch** — two `Const`s can only be hashcons-equal if their payload types match, and checking this first avoids invoking a throwing `isequal`; and
- coerces any non-`Bool` result to `false` (safe for hashconsing — such payloads are simply stored as distinct entries).

Common numeric hashconsing is unaffected: those comparisons go through the generated scalar-pair branches (all 13×13 combinations of the listed `SCALARS`), not the fallback. Verified that e.g. `isequal(Const(1), Const(1.0))` still holds.

## Testing

- Added a regression testset in `test/hash_consing.jl` using a mock symbolic-scalar type with non-`Bool` `isequal`, covering both the direct `isequal_somescalar` call and the end-to-end `substitute(...; fold = Val(true))` path that MTK uses.
- Full `GROUP=Core` suite passes locally on Julia 1.12.4: **17834 pass, 0 fail** (3 pre-existing broken), including the fuzz group.

Patch version bump `4.38.1 → 4.38.2` (bugfix, no public API change).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
-version = "4.38.1"
+version = "4.38.2"
 authors = ["Shashi Gowda"]
 
 [deps]
--- a/src/hashconsing.jl
+++ b/src/hashconsing.jl
@@ -23,8 +23,27 @@ macro __generate_isequal_somescalar()
         cur_expr = new_expr
     end
 
-    push!(cur_expr.args, :(isequal(a, b)::Bool))
+    push!(cur_expr.args, :(isequal_somescalar_fallback(a, b)))
     return esc(expr)
+end
+
+"""
+    $TYPEDSIGNATURES
+
+Fallback comparison for `Const` payloads that are not among the common scalar types.
+
+Some symbolic-scalar payloads (e.g. `CasADi.MX`) break the `isequal` contract: their
+comparison operators are themselves symbolic and return a non-`Bool`, so a plain
+`isequal(a, b)::Bool` throws `TypeError: non-boolean (...) used in boolean context`.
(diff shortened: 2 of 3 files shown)


AayushSabharwal:
Usually if this happens it's a bug in MTK

AayushSabharwal:
Some substitution gone awry.

PR #1003, opened 2026-07-04
🤖 Complex fixes

KristofferC:
I saw a slack thread with some issues w.r.t complex numbers. This is what the robot identified.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -521,6 +521,21 @@ function Base.imag(s::BasicSymbolic{T}) where {T}
         _ => Term{T}(imag, ArgsT{T}((s,)); type = Real)
     end
 end
+function Base.complex(a::BasicSymbolic{T}, b::BasicSymbolic{T}) where {T}
+    Ta = symtype(a)
+    Tb = symtype(b)
+    if Ta <: Real && Tb <: Real
+        if isconst(a) && isconst(b)
+            return Const{T}(complex(unwrap_const(a), unwrap_const(b)))
+        end
+        type = promote_symtype(complex, Ta, Tb)
+        return Term{T}(complex, ArgsT{T}((a, b)); type, shape = ShapeVecT())
+    else
+        # `complex(a, b) == a + im*b` even when the arguments are themselves
+        # complex, e.g. after substituting a complex value into a real slot
+        return a + Const{T}(im) * b
+    end
+end
 
 function promote_symtype(::typeof(adjoint), T::TypeT)
     if T <: Number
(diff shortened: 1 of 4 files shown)


src/methods.jl
```diff
@@ -521,6 +521,21 @@ function Base.imag(s::BasicSymbolic{T}) where {T}
...
+        if isconst(a) && isconst(b)
+            return Const{T}(complex(unwrap_const(a), unwrap_const(b)))
+        end
+        type = promote_symtype(complex, Ta, Tb)
+        return Term{T}(complex, ArgsT{T}((a, b)); type, shape = ShapeVecT())
+    else
+        # `complex(a, b) == a + im*b` even when the arguments are themselves
+        # complex, e.g. after substituting a complex value into a real slot
+        return a + Const{T}(im) * b
+    end
```
AayushSabharwal, line 537: This should error. Substituting a complex value into a real slot is probably not a good idea, and I'd like to avoid having different behavior from `Base` unless absolutely necessary. Constant-folding in `substitute` should also be done via the `fold = Val(true)` kwarg.

src/methods.jl
```diff
@@ -1705,9 +1720,20 @@ function promote_symtype(::typeof(complex), T::TypeT)
...
 end
 function promote_symtype(::typeof(complex), T::TypeT, S::TypeT)
-    @assert T <: Real
-    @assert S <: Real
-    Complex{promote_type_fast_path(T, S)}
+    @assert T <: Number
+    @assert S <: Number
+    # `complex(a, b) == a + im*b`, so complex arguments are also valid
+    rT = T <: Complex ? T.parameters[1]::TypeT : T
+    rS = S <: Complex ? S.parameters[1]::TypeT : S
```
AayushSabharwal, line 1727: Similarly for not allowing complex arguments to `complex`.

src/substitute.jl
```diff
@@ -208,6 +208,11 @@ end
 
 function combine_fold(::Type{T}, op, args::Union{ROArgsT{T}, ArgsT{T}}, meta::MetadataT, can_fold::Bool) where {T}
     @nospecialize op args meta
+    if op === complex
+        # `complex(a, b)` is not defined for non-real `a`/`b`; `maketerm` lowers
+        # such calls to `a + im*b` and folds constants itself
+        return maketerm(BasicSymbolic{T}, op, args, meta)::BasicSymbolic{T}
+    end
```
AayushSabharwal, line 215: This shouldn't be necessary.

src/terminterface.jl
```diff
@@ -342,6 +342,12 @@ function TermInterface.maketerm(::Type{BasicSymbolic{T}}, f, args, metadata; @no
                 @set! res.metadata = metadata
             end
             return res
+        elseif f === complex && length(args) == 2
+            res = complex(args[1], args[2])
+            if metadata !== nothing && iscall(res)
+                @set! res.metadata = metadata
+            end
+            return res
```
AayushSabharwal, line 350: If `complex` doesn't try to constant fold complex arguments manually, this won't be required.

AayushSabharwal (changes requested):
IMO `complex` shouldn't accept complex arguments. Otherwise generating code for such an expression will result in a function that can't be evaluated in Julia.

PR #1006, opened 2026-07-10
Make Mapper and Mapreducer public API

ChrisRackauckas-Claude:
> [!NOTE]
> This PR should be ignored until reviewed by @ChrisRackauckas.

Tracing `map`/`sum`/`prod`/`mapreduce` over symbolic arrays produces terms whose `operation` is a `Mapper`/`Mapreducer` callable. Downstream packages that analyze symbolic expressions need to dispatch on these types to recognize array reductions — the immediate consumer is SymbolicAnalysis.jl, where classifying the curvature of `sum(x)` requires recognizing `Mapreducer(identity, Base.add_sum, ...)` as a plain sum (today such expressions silently analyze as unknown curvature because the operation is never `sum` itself).

This declares both types `@public` with docstrings and adds them to the API reference page, so that dependency is on documented public API rather than internals. No behavior change.

Test suite run locally: 17829 pass, 3 pre-existing broken.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -20,6 +20,12 @@ SymbolicUtils.setmetadata
 SymbolicUtils.promote_symtype
 ```
 
+### Array Operation Callables
+```@docs
+SymbolicUtils.Mapper
+SymbolicUtils.Mapreducer
+```
+
 ## Rewriting System
 
 ### Rule Creation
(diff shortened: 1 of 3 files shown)


ChrisRackauckas-Claude:
The four failing Downstream jobs (ModelingToolkit.jl/InterfaceII, Catalyst.jl/Modeling, Symbolics.jl/GroebnerExt, ModelOrderReduction.jl/Core) fail identically on unmodified master ([run 29078277194](https://github.com/JuliaSymbolics/SymbolicUtils.jl/actions/runs/29078277194)) — pre-existing, not introduced by this PR. A separate investigation of the master failures is underway and will be reported independently.

Issue #1008, opened 2026-07-13
Track the Runic 1.7 formatting backlog


ChrisRackauckas-Claude:
## Summary

Runic 1.7.0 cannot currently complete a clean formatting check of SymbolicUtils. On clean `master` at `1d59a93ede44a7f54e92997b3ea53563cfeb957d` with Julia 1.12.6, per-file checks over all 75 tracked Julia files produced:

- 6 files already formatted
- 68 files with ordinary formatting diffs
- 1 file (`test/new_code.jl`) that aborts Runic while it reparses its output

After replacing only that parser-triggering expected-AST syntax with an equivalent explicit `Expr` construction, Runic reaches all 75 files and reports 69 files with formatting diffs. The formatting backlog is therefore independent of the parser workaround and should be handled separately rather than folded into that focused fix.

## Reproduction

```sh
git checkout 1d59a93ede44a7f54e92997b3ea53563cfeb957d
julia +1.12 --project=/path/to/runic-1.7-environment -m Runic --check --verbose .
```

The whole-repository command aborts at `test/new_code.jl`, so I also ran the same Runic 1.7.0 `--check` command once per path returned by:

```sh
rg --files -g '*.jl' | sort
```

The six files that pass unchanged are:

- `src/despecialize.jl`
- `src/graph_traversal.jl`
- `src/hashconsing.jl`
- `src/stackdict.jl`
- `src/symbolic_codegen_primitives.jl`
- `test/hash_consing.jl`

Every other tracked Julia file needs formatting once the parser-triggering case is represented safely.

## Scope

This issue tracks the 69-file mechanical formatting backlog only. The parser-triggering test expression is being handled as a separate focused change. No formatting sweep was applied as part of that fix.

ChrisRackauckas-Claude:
The parser-triggering test expression is isolated in draft PR #1009. That focused change lets Runic reach all 75 Julia files; the 69-file formatting backlog tracked here remains unchanged.

Issue #186, opened 2021-01-28
Multiple dispatch for symbolic arguments? /lazy evaluation


MilesCranmer:
Hey,

Is there any way I can work with a pre-defined function and a symbolic version at the same time, without having to expand the function for symbolic input?

For example, say I have this function:

```julia
function cube(x)
    x^3
end
```

Now, for simplification reasons, I don't want to expand this. So when I declare:
```julia
@syms x1 x2 cube(x)
f = cube(x1) + x2 + x2
```
and then simplify `f`, I would like the `cube` to remain:

```julia
cube(x1) + 2*x2
```
rather than
```julia
x1^3 + 2*x2
```
(or having it throw an error when I declare it with `@syms`) because then I can evaluate the simplified version on numerical data in SymbolicRegression.jl.

Is this possible? I guess I could set up a manual symbol operator mapping for certain operators, but maybe there's a cleaner built-in way to do this?

Thanks!
Miles

shashi:
This is accomplished by `@register` in [ModelingToolkit and going forward, in Symbolics.jl](https://github.com/SciML/ModelingToolkit.jl/blob/master/test/function_registration.jl#L77)

Basically you will have to define a method of `cube` on symbolic input (your type) to return a symbolic all to the function `cube`.

MilesCranmer:
Thanks @shashi! Will check this out.

Cheers,
Miles

PR #1009, opened 2026-07-13
Avoid ambiguous interpolated tuple syntax in codegen test

ChrisRackauckas-Claude:
> [!IMPORTANT]
> Please ignore this draft until it has been reviewed by @ChrisRackauckas.

## Summary

Replace one interpolated singleton-tuple function signature in the code-generation test's expected AST with the equivalent explicit `Expr` construction.

The test still compares the complete generated function representation. This only avoids source syntax that Runic 1.6 and newer rewrites into an unparsable function signature.

## Root cause and investigation

- Clean `master` at `1d59a93ede44a7f54e92997b3ea53563cfeb957d` reproduced `Runic.AssertionError: re-parsing the formatted output failed` in `test/new_code.jl` with Julia 1.12.6 and Runic 1.7.0.
- A five-line reproducer is a quoted `function ($c,) ... end` expression. The trailing comma is required for Julia to parse the interpolated value as a one-argument anonymous function signature, but Runic removes it and then cannot reparse its output.
- A SymbolicUtils history bisect found `6f61bfd305a48bfacef97fbd9757483ebe19b64b` (`test: test new codegen`) as the first source commit containing the trigger.
- Runic 1.5.1 formats the reproducer, while 1.6.0, 1.6.1, and 1.7.0 fail. A Runic source bisect found `c89356cc36aa52798eeef6f2a484171add95751d` (`Upgrade to JuliaSyntax version 1 (#186)`) as the first bad formatter commit.
- Constructing the signature as `Expr(:tuple, c)` is structurally equal to the original parsed signature and avoids relying on that ambiguous surface syntax.

No Runic issue or PR was opened because that repository is outside the authorized scope for this task.

## Validation

- Direct Julia 1.12.6 comparison after `Base.remove_linenums!`: original quote AST and explicit `Expr` AST are equal and have identical `repr` output.
- `GROUP=Core julia +1.12 --project=. -e 'using Pkg; Pkg.test()'`: 17,830 passed, 8 pre-existing broken, 17,838 total in 14m19.9s.
- Runic 1.7.0 on the exact changed range (`--lines=347:350 --inplace`): exit 0 and the file's Git blob hash remained unchanged (`de4bc35cc21e4c46f8e7fca9204fc924392dc8c1`).
- Whole-repository Runic 1.7.0 now reaches all 75 Julia files without an assertion. It still exits 1 for the independent pre-existing formatting backlog: 6 formatted and 69 unformatted files.
- `git diff --check`: clean.

The separate mechanical formatting backlog is tracked in #1008. This PR intentionally does not reformat the other files.

### DIFF
--- a/test/new_code.jl
+++ b/test/new_code.jl
@@ -344,8 +344,10 @@ test_repr(a, b) = @test repr(Base.remove_linenums!(a)) == repr(Base.remove_linen
 
     test_repr(
         Code.fast_toexpr(Func([DestructuredArgs([a, b], c, inds = [:a, :b])], [], a + b), ir, Dict()),
-        :(
-            function ($c,)
+        Expr(
+            :function,
+            Expr(:tuple, c),
+            quote
                 begin
                     __miscₛᵧₘ0 = c.a
                     a = __miscₛᵧₘ0


ChrisRackauckas-Claude:
Investigation scratchpad for `7c28480230af52d35bbdab0c69f11e2f7421fdcf`:

1. Synced the fork and created the feature branch from exact upstream `1d59a93ede44a7f54e92997b3ea53563cfeb957d`.
2. Reproduced the clean-main Runic 1.7.0 assertion with Julia 1.12.6 in a workspace-local environment.
3. Reduced it to:

   ```julia
   quote
       function ($c,)
           nothing
       end
   end
   ```

4. Bisected SymbolicUtils history to `6f61bfd305a48bfacef97fbd9757483ebe19b64b` and Runic history to `c89356cc36aa52798eeef6f2a484171add95751d`.
5. Confirmed the Runic release boundary: 1.5.1 passes; 1.6.0, 1.6.1, and 1.7.0 fail.
6. Compared the parsed original and explicit-constructor ASTs after removing line numbers. Both structural equality and `repr` equality passed.
7. Ran the native full `Core` group: 17,830 pass / 8 pre-existing broken / 17,838 total.
8. Ran Runic 1.7.0 on the exact changed lines. It exited 0 and left the file hash unchanged.
9. Re-ran whole-repository Runic. The parser assertion is gone; the separate 69-file mechanical formatting backlog remains and is recorded in #1008.

No tests were skipped, disabled, loosened, or silenced.

PR #1014, opened 2026-07-21
Memoize scalarize to share sub-expressions

### LINKED ISSUE #1015: scalarize is super-linear on nested/reused symbolic matrix expressions (no sub-expression sharing)
## Summary

`SymbolicUtils.scalarize` re-expands each shared sub-expression **once per
occurrence** rather than once. `BasicSymbolic` is hash-consed, so an expression
built by reusing a matrix (e.g. an orthonormal frame, a rotation matrix) at
several nesting levels is a *DAG* with heavy structural sharing — but the
`scalarize` recursion walks it as a *tree*. The cost then grows roughly ×4 in
expression size per nesting level and can reach minutes on expressions that are
conceptually small.

This is very easy to hit with any corotational / rotation-matrix formulation,
where one reused frame appears many times in the final equations.

## MWE

```julia
using SymbolicUtils, LinearAlgebra
using SymbolicUtils: scalarize

@syms x[1:3]::Real R[1:3, 1:3]::Real
xv = scalarize(x); Rm = scalarize(R)

"Orthonormal element frame with the sqrt/division of a real corotational beam."
function element_frame(d, R)
    len = sqrt(sum(abs2, d))
    e1 = d ./ len
    y = R[:, 2]
    e2raw = y .- sum(y .* e1) .* e1
    e2 = e2raw ./ sqrt(sum(abs2, e2raw))
    e3 = e1 × e2
    return [e1[1] e2[1] e3[1]; e1[2] e2[2] e3[2]; e1[3] e2[3] e3[3]]
end

"Reuse the frame `depth` extra times, as a beam element nests it."
function nested_out(E, R, depth)
    M = E
    for _ in 1:depth
        M = (E' * R) * M
    end
    theta = [0.5*(M[3,2]-M[2,3]), 0.5*(M[1,3]-M[3,1]), 0.5*(M[2,1]-M[1,2])]
    F = [theta[1], theta[2]+theta[3], theta[2]-theta[3]]
    return E * F
end

E = element_frame(xv, Rm)
for depth in 0:3
    out = nested_out(E, Rm, depth)
    scalarize.(out)                       # warm up
    t = @elapsed scalarize.(out)
    println("depth=$depth  scalarize=$(round(t; digits=4))s")
end
```

### Measured (Julia 1.12, SymbolicUtils master v4.40.2)

| depth | `scalarize` |
|-------|-------------|
| 0     | 0.024 s     |
| 1     | 0.274 s     |
| 2     | 1.046 s     |
| 3     | 3.158 s     |

Time grows super-linearly with nesting depth even though only one small frame is
being reused. Real corotational-beam formulations nest deeper than this and hit
multi-minute `scalarize` calls.

## Cause

`scalarize(x::BasicSymbolic, ::Val{toplevel})` recurses into the sub-expressions
of `x` with no memoization. When the same sub-object is reached through many
paths (which is the common case, since `BasicSymbolic` is hash-consed), it is
fully re-scalarized every time.

## Proposed fix

Memoize `scalarize` within a single top-level call using an identity-keyed
(`IdDict`) cache: because shared sub-terms are the *same object*, each unique
sub-expression is expanded exactly once. Scope the cache to one top-level call
(via `ScopedValue`) so the memory is released afterwards, and copy array results
on return so callers can't alias/mutate the cached value.

Implemented in #1014, which is ~1000× faster at nesting depth 3 in the MWE above
with an identical (structurally `hash`-equal) result.

## Version info

- Julia 1.12
- SymbolicUtils master (v4.40.x)

1-Bart-1:
Fixes #1015

## What

`scalarize(x::BasicSymbolic, ::Val{toplevel})` recursed into sub-expressions with no memoization, so a shared sub-expression was re-expanded **once per occurrence**. Because `BasicSymbolic` is hash-consed, an expression that reuses one matrix (a rotation matrix, an orthonormal frame) at several nesting levels is a DAG with heavy sharing — but the recursion walked it as a tree, giving super-linear (~×4 in expression size per nesting level) blow-up. Any corotational / rotation-matrix formulation hits this easily.

This PR adds a per-call, identity-keyed memo at the `scalarize` recursion hub:

- The cache is an `IdDict` held in a `ScopedValue`, established on the first (top-level) `scalarize` entry and torn down when that call returns, so nothing is retained between independent `scalarize` calls.
- Because shared sub-terms are the *same object*, an `IdDict` lookup expands each unique sub-expression exactly once.
- Only non-`toplevel` results are cached and read (the `toplevel` path folds constants and has different semantics, so it is never served from or written to the cache).
- Array results are `copy`d on return (`scalarize_uncache`) so a caller that mutates the returned array in place cannot corrupt the cached master.

The scalarization logic itself is unchanged — it was extracted verbatim into `scalarize_impl`, and the public `scalarize` now wraps it with the cache check.

## Correctness

The scalarized output is the same expression as before — in the MWE below the structural `hash` of the result is identical on `master` and this branch at every nesting depth. A regression test (`test/arrayop.jl`, "scalarize shares reused sub-expressions") asserts a depth-3 corotational-frame expression scalarizes to the correct numeric result.

## Benchmark (Julia 1.12, MWE below)

| depth | master  | this PR  | speedup |
|-------|---------|----------|---------|
| 0     | 0.024 s | 0.0009 s | ~26×    |
| 1     | 0.274 s | 0.0021 s | ~131×   |
| 2     | 1.046 s | 0.0028 s | ~373×   |
| 3     | 3.158 s | 0.0031 s | ~1019×  |

## MWE

```julia
using SymbolicUtils, LinearAlgebra
using SymbolicUtils: scalarize

@syms x[1:3]::Real R[1:3, 1:3]::Real
xv = scalarize(x); Rm = scalarize(R)

function element_frame(d, R)
    len = sqrt(sum(abs2, d))
    e1 = d ./ len
    y = R[:, 2]
    e2raw = y .- sum(y .* e1) .* e1
    e2 = e2raw ./ sqrt(sum(abs2, e2raw))
    e3 = e1 × e2
    return [e1[1] e2[1] e3[1]; e1[2] e2[2] e3[2]; e1[3] e2[3] e3[3]]
end

function nested_out(E, R, depth)
    M = E
    for _ in 1:depth
        M = (E' * R) * M
    end
    theta = [0.5*(M[3,2]-M[2,3]), 0.5*(M[1,3]-M[3,1]), 0.5*(M[2,1]-M[1,2])]
    F = [theta[1], theta[2]+theta[3], theta[2]-theta[3]]
    return E * F
end

E = element_frame(xv, Rm)
for depth in 0:3
    out = nested_out(E, Rm, depth)
    scalarize.(out)                       # warm up
    t = @elapsed sc = scalarize.(out)
    println("depth=$depth  scalarize=$(round(t; digits=4))s  hash=$(hash(sc))")
end
```


AayushSabharwal (approved):

AayushSabharwal:
Ah, ScopedValue doesn't exist on 1.10. This probably needs https://github.com/JuliaLang/ScopedValues.jl/ then.

AayushSabharwal:
It also looks like some SnoopCompile thing is breaking the rest of CI

1-Bart-1:
snoopcompile wasn't actually used, so I removed the dep. Using the ScopedValues pkg instead now.

1-Bart-1:
another fix for an actual regression pushed now

AayushSabharwal:
Thank you for your persistence here. The 1.10 failure is weird. As far as I can tell, the matmul code works differently on newer Julia versions, and the 1.10 version just doesn't like symbolic arrays.

AayushSabharwal:
```julia
Base.copy(x::BasicSymbolic{T}) where {T} = x
```
Might help

KristofferC:
- I don't think the ScopedValue usage is motivated here (`IdDict` is not thread-safe, so you still have a race if `scalarize` would spawn a task that mutates the dict). You can use a `TaskLocalValue` and reset it in the end. And why can't we use `@cache` like for the other caches?
- Related to the `@cache` question, things like `scalarize.(out)` will use a fresh cache per element, seems wasteful?

AayushSabharwal:
> why can't we use @cache like for the other caches?

`@cache` doesn't copy the cached value, and the user is free to mutate the scalarized array.

KristofferC:
But that is true for this implementation as well; that's why the `_uncache` functions are there? I guess the question is how bad it is to retain these values between calls (will they blow up the cache size etc).

AayushSabharwal:
This can use a `TaskLocalValue` for the cache, I agree. And it should be thread-safe. The `_uncache` phrasing is necessary because this implementation caches `result` and returns `..._uncache(result)`. `@cache` will cache whatever is returned from the function, and return the same thing to the user. The control is lost.

KristofferC:
> @cache will cache whatever is returned from the function, and return the same thing to the user. The control is lost.

But can't you just put the cache retrieval in an aux function

```
foo(x) = copy(_foo_cached(x))
@cache _foo_cached(x) = ....
```

AayushSabharwal:
Oh, right I didn't think of that. That would work.

1-Bart-1:
Now without ScopedValue

KristofferC:
So the remaining question now is, how much and how big stuff will this cache? Does it need a lower cache `limit` than default?

AayushSabharwal:
It's hard to say. The default limit is 100_000. The size of the elements cached depends completely on what is being scalarized.

1-Bart-1:
For a full cache the dict itself is only a couple of MB, and I see around 90MB including the symbolic expressions. If someone needs a tighter cap they can use `set_limit!`.

1-Bart-1:
Anything I can do to make this mergeable?

AayushSabharwal:
`Base.summarize` on the cache dict for a reasonably large Multibody model is just 7MB, and this includes entire expression trees which would otherwise consume memory anyway. I think we're good on the memory usage front. Claude also gave me these statistics:
```julia
┌────────────────────────────┬────────────┬───────┬────────┬──────────┐
│           Stage            │ Cache size │ Hits  │ Misses │ Hit rate │
├────────────────────────────┼────────────┼───────┼────────┼──────────┤
│ After multibody(m1)        │ 2,484      │ 2,660 │ 2,484  │ 51.7%    │
├────────────────────────────┼────────────┼───────┼────────┼──────────┤
│ After full ODEProblem(...) │ 3,343      │ 7,781 │ 3,343  │ ~70.0%   │
└────────────────────────────┴────────────┴───────┴────────┴──────────┘
```
Which is pretty good. Thanks @1-Bart-1, I'll merge.

PR #1052, opened 2026-09-04
Compare an expression as the DAG it is, not the tree it denotes

### LINKED ISSUE #1049: isequal on two equal-but-distinct expressions is exponential in nesting depth
`isequal` on two structurally equal but non-identical `BasicSymbolic` values walks the
expression as a **tree** rather than the DAG it is: a shared subterm is compared once per
path that reaches it, not once in total. That is `O(2^depth)` on a graph with `O(depth)`
nodes.

It is reachable from ordinary use — hash-consing calls `isequal` on a cache hit, and the
per-key lookups inside `isequal_addmuldict` run with `COMPARE_FULL = false`, where the
id-based shortcut cannot fire.

## Reproducer

```julia
using SymbolicUtils

struct Tag end
@syms x::Real y::Real

# Depth-n DAG: `e` is shared by both factors, so the tree it denotes is
# exponentially larger than the graph it is stored as.
function deep(leaf, n)
    e = leaf
    for i in 1:n
        e = (e + i) * (e + y)
    end
    return e
end

# Same structure, different leaf metadata, so hash-consing keeps these as two
# distinct objects.
tagged = SymbolicUtils.setmetadata(x, Tag, 1)

for n in 16:24
    println("depth=", n, "  isequal = ", @elapsed(isequal(deep(x, n), deep(tagged, n))), " s")
end
```

SymbolicUtils v4.46.2, Julia 1.12.7 — each level roughly doubles the time, on a graph of
about `2n` nodes:

```
depth=16  0.0347 s
depth=18  0.1745 s
depth=20  0.7089 s
depth=22  1.9933 s
depth=24  8.0718 s
```

## Cause

The id-based shortcut in `isequal_bsimpl` is guarded on `full`:

```julia
if full && ida !== idb && ida !== nothing && idb !== nothing
    return false
end
```

That guard is correct — under `COMPARE_FULL = false` metadata is ignored, so two values
with different ids can legitimately be equal. What is missing is memoisation: nothing
records which `(a, b)` pairs have already been compared, so subterms are re-visited once
per path.

## Impact

Building a ModelingToolkit model, then an equivalent one in the same session, made
`ODEProblem` construction take **320 s** against **2.2 s** for the first build
(`mtkcompile` was unaffected). In CI, with ~10 models built in one process first, it
reached **4460 s**; the same model in a fresh process takes 1.2 s. Instrumenting
`isequal_addmuldict`, the worst single comparison in that build spawned **19 011 535**
nested calls — from a **two-entry** dictionary, at `AddMul` nesting depth 23
(`2^24 ≈ 19e6`). Summand count is irrelevant; depth is everything.

## Fix

PR #1052 adds a task-local memo of visited `(a, b, full)` pairs, opened by the outermost
`Base.isequal` and cleared on exit. Depth 24 goes from 8.07 s to 2.0e-5 s and scaling
becomes linear (depth 200: 1.97e-4 s). `Pkg.test("SymbolicUtils")` passes and equality
answers are unchanged.

1-Bart-1:
Fixes #1049.

`isequal` on two structurally equal but non-identical `BasicSymbolic` values walked the
expression as a tree rather than the DAG it is: a subterm reachable by several paths was
compared once per path instead of once in total. That is `O(2^depth)` on a graph with
`O(depth)` nodes.

It is reachable from ordinary use — hash-consing calls `isequal` on a cache hit, and the
per-key lookups inside `isequal_addmuldict` run with `COMPARE_FULL = false`, where the
id-based shortcut cannot fire.

### The guard is not the bug

`isequal_bsimpl` already has a shortcut that would cut the recursion for two distinct
interned nodes:

```julia
if full && ida !== idb && ida !== nothing && idb !== nothing
    return false
end
```

I have deliberately left it as it is. Under `COMPARE_FULL = false` the comparison ignores
metadata, so two values with different ids can legitimately be equal under that relation,
and concluding otherwise would be wrong. What was missing is memoisation, not a wider
shortcut.

### The change

A task-local `Dict{Tuple{UInt, UInt, Bool}, Bool}` memo, keyed by the two objects'
identities and the current `full` flag. The outermost `Base.isequal` creates it and
clears it in a `finally`, so nested comparisons share one memo and no `objectid` key
outlives the object it identifies.

### Effect

Reproducer from #1049 — a depth-`n` DAG of about `2n` nodes:

| depth | before | after |
| ----- | -------- | -------- |
| 16 | 0.0347 s | 2.1e-5 s |
| 20 | 0.7089 s | 2.0e-5 s |
| 24 | 8.0718 s | 2.0e-5 s |
| 100 | — | 1.07e-4 s |
| 200 | — | 1.97e-4 s |

Doubling the depth now doubles the time rather than squaring it.

Downstream, this is what made `ODEProblem` construction take 320 s for a model that
builds in 2 s in a fresh session, and 4460 s in CI — the second and later builds of
equivalent models hit the cache and paid the exponential walk on every reconstructed
node.

### Checks

- `Pkg.test("SymbolicUtils")` passes.
- Equality answers are unchanged, including the negative cases (different leaf, different
  depth, and the identical-object fast path).

### DIFF
--- a/src/hashconsing.jl
+++ b/src/hashconsing.jl
@@ -9,6 +9,18 @@ information.
 """
 const COMPARE_FULL = TaskLocalValue{Bool}(Returns(false))
 
+"""
+Task-local memo of `(objectid(a), objectid(b), full) => result` for the comparison
+currently in progress, or `nothing` when none is running.
+
+Expressions are DAGs, so a subterm reachable by several paths would otherwise be
+compared once per path — exponential in the nesting depth on a graph linear in it. The
+entry point in `Base.isequal` opens the memo and closes it again, so a key cannot
+outlive the objects whose identity it records.
+"""
+const EQUALITY_MEMO =
+    TaskLocalValue{Union{Nothing, Dict{Tuple{UInt, UInt, Bool}, Bool}}}(Returns(nothing))
+
 macro __generate_isequal_somescalar()
     expr = Expr(:if)
     cur_expr = expr
@@ -194,6 +206,13 @@ function isequal_bsimpl(a::BSImpl.Type{T}, b::BSImpl.Type{T}, full::Bool) where 
         return false
     end
 
+    memo = EQUALITY_MEMO[]
+    memo_key = (objectid(a), objectid(b), full)
+    if memo !== nothing
+        cached = get(memo, memo_key, nothing)
+        cached === nothing || return cached
+    end
+
     partial = @match (a, b) begin
         (BSImpl.Const(; val = v1), BSImpl.Const(; val = v2)) => begin
             isequal_somescalar(v1, v2)::Bool && (!full || (typeof(v1) === typeof(v2))::Bool)
@@ -220,6 +239,7 @@ function isequal_bsimpl(a::BSImpl.Type{T}, b::BSImpl.Type{T}, full::Bool) where 
     if full && partial && !(Ta <: BSImpl.Const)
         partial = metadata_isequal(metadata(a), metadata(b))
     end
+    memo === nothing || (memo[memo_key] = partial)
     return partial
 end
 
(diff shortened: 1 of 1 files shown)


oameye:
This introduced a performance regression in QuantumCumulants.jl. Tried to fix it in #1063

ChrisRackauckas:
Interesting this could give a regression... it should be strictly faster 😅  but I'll check.

oameye:
Thanks for the swift fix!

PR #1058, opened 2026-09-06
Compare copied symbolic identities structurally during interning

ChrisRackauckas:
Re-applies #1057 onto current master; it conflicted once #1056 landed, since both edit `src/hashconsing.jl` and `test/hash_consing.jl`.

Distinct intern IDs are no longer treated as proof of inequality during full symbolic comparison. `deepcopy` mints a new mutable ID token without changing the expression, so that shortcut stopped `sin(deepcopy(x))` from reusing the canonical expression and grew collision chains. The same-ID positive shortcut is kept; different IDs fall through to the existing structural and metadata comparison.

The source change is unchanged from #1057 — the removed block is still present on master. Only the test insertion point moved.

https://claude.ai/code/session_014FEzNTLFutCmTEAZ3zBg5R

### DIFF
--- a/src/hashconsing.jl
+++ b/src/hashconsing.jl
@@ -203,10 +203,6 @@ function isequal_bsimpl(a::BSImpl.Type{T}, b::BSImpl.Type{T}, full::Bool)::Bool 
     Tb = MData.variant_type(b)
     Ta === Tb || return false
 
-    if full && ida !== idb && ida !== nothing && idb !== nothing
-        return false
-    end
-
     memo = EQUALITY_MEMO[]
     memo_key = (objectid(a), objectid(b), full)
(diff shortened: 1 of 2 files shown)


src/hashconsing.jl
```diff
@@ -203,10 +203,6 @@ function isequal_bsimpl(a::BSImpl.Type{T}, b::BSImpl.Type{T}, full::Bool)::Bool
     Tb = MData.variant_type(b)
     Ta === Tb || return false
 
-    if full && ida !== idb && ida !== nothing && idb !== nothing
```
AayushSabharwal, line 206: See #1057

AayushSabharwal (changes requested):

PR #1055, opened 2026-09-05
Make DAG equality memoization identity-safe and reuse task-local storage

ChrisRackauckas-Claude:
Follow up on https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/1052: an identity-hash collision can make `isequal` return true for unequal expressions. Memo keys now retain both objects and compare them with `===`, including the comparison mode. Task-owned storage is reused and emptied in `finally`; a small buffer avoids clearing a table enlarged by an earlier DAG comparison on every subsequent small comparison.

Adds regressions for collisions in both storage paths, linear comparison counts on shared DAGs, separate metadata modes, exception cleanup, and concurrent tasks comparing shared expressions. Cheap name/ID rejection checks run before entering the memo. The identical-object check stays in a small inlined entry point, keeping memo setup and cleanup out of that fast path.

Please ignore this draft until reviewed by @ChrisRackauckas.

Includes the separate documentation prerequisite https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/1054. Those three reference entries fix existing master docs/QA failures and remain separate commits. This PR targets master because the repository's Core/QA workflow only runs for PRs targeting master or v4.

### Failing before / passing after

The same isolated collision regression runs against both checkouts:

```sh
julia +1.12 --compiled-modules=existing --compile=min -O0 --project=collision-master collision_probe.jl
julia +1.12 --compiled-modules=existing --compile=min -O0 --project=SymbolicUtils.jl collision_probe.jl
```

Before, on unmodified master at https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/748075182f48408d6159b1a6e077cbb55921d695:

```text
Identity hash collisions: Test Failed
Expression: !result
Test Summary:              | Pass  Fail  Total
Equality memo              |    2     1      3
  Identity hash collisions |    2     1      3
```

With the fix:

```text
Test Summary: | Pass  Total
Equality memo |    3      3
```

An independent adjacent-commit check confirmed that the immediate parent passes the same probe. Introducing commit: https://github.com/JuliaSymbolics/SymbolicUtils.jl/commit/748075182f48408d6159b1a6e077cbb55921d695.

The probe specializes `objectid` to a constant for a test-only symbolic subtype, a deterministic way to exercise a permitted hash collision. The committed test covers both a short comparison and enough distinct pairs to exercise the dictionary.

<details>
<summary>Standalone collision probe</summary>

```julia
using SymbolicUtils, Test
[...]

PR #1063, opened 2026-09-08
perf: reuse equality memo prototype

oameye:
This fixes the allocation regression introduced by #1052 without removing the DAG-aware equality memoization.

The current implementation creates a new `Dict` for every outer `BasicSymbolic` equality comparison. That is expensive for workloads that perform many ordinary symbolic comparisons, such as QuantumCumulants.jl.

This PR reuses one task-local memo dictionary and clears it between comparisons instead.


oameye:
Okay https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/1055 seems to be the better solution

ChrisRackauckas:
This should be covered in https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/1055

PR #1064, opened 2026-09-08
Truncate Base.hasha_seed to UInt for 32-bit Julia

ChrisRackauckas-Claude:
## Summary
- `Base.hasha_seed` is always `UInt64` (`0x8b4142914d506ccb`).
- `#1060` truncated package hex salts with `% UInt`, but left `hv = Base.hasha_seed` bare.
- On i686 that becomes `InexactError: trunc(UInt32, …)` during `hash_bsimpl` / precompile, cascading into Symbolics, Sundials, ModelingToolkit, etc.
- Bump to **4.46.4** for registration.

## Test plan
- [ ] Existing x86 Core CI green
- [ ] `using SymbolicUtils` precompiles on 32-bit Julia

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
-version = "4.46.3"
+version = "4.46.4"
 authors = ["Shashi Gowda"]
 
 [deps]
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
@JuliaRegistrator register

PR #1067, opened 2026-09-09
fix: publish lazy argument caches atomically

### LINKED ISSUE #696: Hash consing and threading leads to crashes
```julia
using SymbolicUtils

Threads.nthreads()

SymbolicUtils.ENABLE_HASHCONSING[] = false

function testfunc()
    @syms x::Number y::Number f(::Number)::Number
    e1 = f(x) + y
    e2 = f(cos(x)) + e1^2
    e3 = sin(e2)
    substitute(e3, Dict(f(x)=>2))
end

# Threads without hash consing: works!
@Threads.threads for i in 1:100
    testfunc()
end

SymbolicUtils.ENABLE_HASHCONSING[] = true

# Hash consing without threads: works!
empty!(SymbolicUtils.wvd)
for i in 1:100
    testfunc()
end

# Hash consing with threads: exceptions or segfaults! May need to run more than once.
empty!(SymbolicUtils.wvd)
@Threads.threads for i in 1:100
    testfunc()
end
```
I have a guess at what might be going wrong. It looks like the code in `arguments()` [here](https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/4b0d24ef7571c396f192393fda33d5e0ff7de981/src/types.jl#L168) initializes the arguments vector without any locking. Hash consing might mean that two threads with apparently independently-defined symbols with the same content now call `arguments` on the same newly-initialized symbolic struct at the same time, so the arguments list could get clobbered.

An example exception:
```julia
ERROR: TaskFailedException

    nested task error: UndefRefError: access to undefined reference
    Stacktrace:
      [1] getindex
        @ ./essentials.jl:13 [inlined]
      [2] iterate
        @ ./array.jl:945 [inlined]
      [3] _foldl_impl
        @ ./reduce.jl:60 [inlined]
      [4] foldl_impl
        @ ./reduce.jl:48 [inlined]
      [5] mapfoldl_impl(f::typeof(SymbolicUtils.symtype), op::SymbolicUtils.var"#37#38"{…}, nt::Base._InitialValue, itr::Vector{…})
        @ Base ./reduce.jl:44
      [6] mapfoldl(f::Function, op::Function, itr::Vector{SymbolicUtils.BasicSymbolic{Number}}; init::Base._InitialValue)
        @ Base ./reduce.jl:175
      [7] mapfoldl
        @ ./reduce.jl:175 [inlined]
      [8] _promote_symtype(f::Function, args::Vector{SymbolicUtils.BasicSymbolic{Number}})
        @ SymbolicUtils ~/.julia/packages/SymbolicUtils/2cE5G/src/types.jl:1181
      [9] maketerm(T::Type{…}, head::Function, args::Vector{…}, metadata::Nothing)
        @ SymbolicUtils ~/.julia/packages/SymbolicUtils/2cE5G/src/types.jl:777
     [10] substitute(expr::SymbolicUtils.BasicSymbolic{Number}, dict::Dict{SymbolicUtils.BasicSymbolic{…}, Int64}; fold::Bool)
        @ SymbolicUtils ~/.julia/packages/SymbolicUtils/2cE5G/src/substitute.jl:34
     [11] substitute
        @ ~/.julia/packages/SymbolicUtils/2cE5G/src/substitute.jl:16 [inlined]
     [12] (::SymbolicUtils.var"#343#345"{Bool, Dict{…}})(x::SymbolicUtils.BasicSymbolic{Number})
        @ SymbolicUtils ~/.julia/packages/SymbolicUtils/2cE5G/src/substitute.jl:24
     [13] iterate
        @ ./generator.jl:47 [inlined]
     [14] _collect(c::Vector{…}, itr::Base.Generator{…}, ::Base.EltypeUnknown, isz::Base.HasShape{…})
        @ Base ./array.jl:854
     [15] collect_similar
        @ ./array.jl:763 [inlined]
     [16] map
        @ ./abstractarray.jl:3286 [inlined]
     [17] substitute(expr::SymbolicUtils.BasicSymbolic{Number}, dict::Dict{SymbolicUtils.BasicSymbolic{…}, Int64}; fold::Bool)
        @ SymbolicUtils ~/.julia/packages/SymbolicUtils/2cE5G/src/substitute.jl:23
     [18] substitute(expr::SymbolicUtils.BasicSymbolic{Number}, dict::Dict{SymbolicUtils.BasicSymbolic{Number}, Int64})
        @ SymbolicUtils ~/.julia/packages/SymbolicUtils/2cE5G/src/substitute.jl:16
     [19] macro expansion
        @ ./REPL[1]:8 [inlined]
     [20] (::var"#9#threadsfor_fun#12"{var"#9#threadsfor_fun#11#13"{UnitRange{Int64}}})(tid::Int64; onethread::Bool)
        @ Main ./threadingconstructs.jl:215
     [21] #9#threadsfor_fun
        @ ./threadingconstructs.jl:182 [inlined]
     [22] (::Base.Threads.var"#1#2"{var"#9#threadsfor_fun#12"{var"#9#threadsfor_fun#11#13"{UnitRange{Int64}}}, Int64})()
        @ Base.Threads ./threadingconstructs.jl:154

...and 1 more exception.

Stacktrace:
 [1] threading_run(fun::var"#9#threadsfor_fun#12"{var"#9#threadsfor_fun#11#13"{UnitRange{Int64}}}, static::Bool)
   @ Base.Threads ./threadingconstructs.jl:172
 [2] macro expansion
   @ ./threadingconstructs.jl:220 [inlined]
 [3] top-level scope
   @ ./REPL[1]:7
Some type information was truncated. Use `show(err)` to see complete types.
```

### LINKED ISSUE #1066: Concurrent first calls to arguments can race while materializing lazy argument caches
`arguments` lazily materializes and caches argument vectors for some `BasicSymbolic` variants. If multiple tasks call `arguments` concurrently on the same previously-unmaterialized expression, they can mutate the same cache vector concurrently.

A minimal reproducer using only SymbolicUtils is:

```julia
using SymbolicUtils
using SymbolicUtils: SymReal
using TermInterface
using Base.Threads

@assert nthreads() > 1

function trial(k; nargs = 4096, ntasks = max(8, 4nthreads()))
    xs = [
        SymbolicUtils.Sym{SymReal}(Symbol("x_$(k)_$(i)"); type = Real)
        for i in 1:nargs
    ]
    expr = sum(xs)

    ready = Atomic{Int}(0)
    go = Atomic{Bool}(false)

    tasks = [
        @spawn begin
            atomic_add!(ready, 1)
            while !go[]
                yield()
            end
            arguments(expr)
        end for _ in 1:ntasks
    ]

    while ready[] < ntasks
        yield()
    end
    go[] = true

    foreach(fetch, tasks)
end

for k in 1:200
    @info "trial" k
    trial(k)
end
```

Run with, for example:

```bash
julia --threads=8 --project=. repro.jl
```

On current `master` this can fail with:

```text
ConcurrencyViolationError("Vector can not be resized concurrently")
```

The race comes from lazy cache initialization mutating the shared argument vector with `push!`. A caller performing concurrent read-only traversal of a shared symbolic DAG can therefore race during the first `arguments(expr)` access.

The lazy cache itself appears intentional and useful; the issue is the publication mechanism. Ideally the cache should only become visible after a complete argument vector has been constructed, so concurrent readers cannot observe or mutate partially initialized state.

The same lazy mutable-cache pattern is used by `AddMul`, `ArrayOp`, and `ArrayMaker`.

oameye:
Closes #1066.
Closes #696.

This also fixes the SymbolicUtils root cause reported in
JuliaDynamics/NetworkDynamics.jl#384.

Lazy argument caches for `AddMul`, `ArrayOp`, and `ArrayMaker` are currently populated by mutating the cache's `SmallVec` in place. Concurrent first calls to `arguments` on the same hash-consed symbolic node can therefore write to the same vector and fail with a `ConcurrencyViolationError`.

This changes lazy cache initialization to use one-time atomic publication:

1. A reader first checks whether a complete cached `ArgsT` has already been published.
2. On a cache miss, it constructs the argument vector privately.
3. The complete vector is published with an atomic compare-and-swap.
4. If another task won the race, the losing task discards its candidate and returns the already-published vector.

This preserves lazy argument construction and avoids placing a lock on the warm `arguments` path. Readers never mutate a shared partially initialized argument vector.

The change applies only to the three variants with lazily derived argument caches:

- `AddMul`
- `ArrayOp`
- `ArrayMaker`

Ordinary `Term.args` remains unchanged.

`ConstructionBase.setproperties` continues to invalidate the derived argument cache when constructing a modified node.

A regression test synchronizes multiple tasks on the first `arguments(expr)` access and verifies that all readers receive complete arguments backed by the same published cache. The test also handles the normal single-threaded package test configuration by spawning a multithreaded Julia subprocess.

Targeted validation was performed with 8 Julia threads, including repeated synchronized concurrent first-access trials. The atomic cache-holder mechanism was also checked on Julia 1.10 and Julia 1.12.

### DIFF
--- a/src/terminterface.jl
+++ b/src/terminterface.jl
@@ -135,6 +135,15 @@ function __sorted_args(x::BasicSymbolic{T})::ROArgsT{T} where {T}
     end
 end
 
+@inline function _get_cached_arguments(cache::ArgsCacheT{T}) where {T}
+    return @atomic :acquire cache.value
+end
+
+@inline function _publish_cached_arguments!(cache::ArgsCacheT{T}, candidate::ArgsT{T}) where {T}
+    result = @atomicreplace :acquire_release :acquire cache.value nothing => candidate
+    return result.success ? candidate : result.old::ArgsT{T}
+end
+
 """
     arguments(expr)
 
@@ -175,11 +184,14 @@ function TermInterface.arguments(x::BSImpl.Type{T})::ROArgsT{T} where {T}
         BSImpl.Sym(_) => throw(ArgumentError("`Sym` does not have arguments."))
         BSImpl.Term(; args) => ROArgsT{T}(args)
         BSImpl.AddMul(; coeff, dict, variant, args, shape, type) => begin
-            isempty(args) || return ROArgsT{T}(args)
+            cached = _get_cached_arguments(args)
+            cached === nothing || return ROArgsT{T}(cached)
+
+            newargs = ArgsT{T}()
             @match variant begin
                 AddMulVariant.ADD => begin
                     if !iszero(coeff)
-                        push!(args, Const{T}(coeff))
+                        push!(newargs, Const{T}(coeff))
                     end
                     for (k, v) in dict
                         newterm = @match k begin
@@ -188,39 +200,45 @@ function TermInterface.arguments(x::BSImpl.Type{T})::ROArgsT{T} where {T}
                             end
                             _ => Mul{T}(v, ACDict{T}(k => 1); shape, type)
                         end
-                        push!(args, newterm)
+                        push!(newargs, newterm)
                     end
                 end
                 AddMulVariant.MUL => begin
                     if !_isone(coeff)
-                        push!(args, Const{T}(coeff))
+                        push!(newargs, Const{T}(coeff))
                     end
                     for (k, v) in dict
-                        push!(args, k ^ v)
+                        push!(newargs, k ^ v)
                     end
                 end
             end
-            return ROArgsT{T}(args)
+            return ROArgsT{T}(_publish_cached_arguments!(args, newargs))
         end
         BSImpl.Div(num, den) => ROArgsT{T}(ArgsT{T}((num, den)))
         BSImpl.ArrayOp(; output_idx, expr, reduce, term, ranges, shape, type, args) => begin
             if term === nothing
-                isempty(args) || return ROArgsT{T}(args)
-                push!(args, Const{T}(output_idx))
-                push!(args, Const{T}(expr))
-                push!(args, Const{T}(reduce))
-                push!(args, Const{T}(term))
-                push!(args, Const{T}(ranges))
-                return ROArgsT{T}(args)
+                cached = _get_cached_arguments(args)
+                cached === nothing || return ROArgsT{T}(cached)
+
+                newargs = ArgsT{T}()
+                push!(newargs, Const{T}(output_idx))
+                push!(newargs, Const{T}(expr))
+                push!(newargs, Const{T}(reduce))
+                push!(newargs, Const{T}(term))
+                push!(newargs, Const{T}(ranges))
+                return ROArgsT{T}(_publish_cached_arguments!(args, newargs))
             elseif term isa BasicSymbolic{T}
                 return arguments(term)
             end
         end
         BSImpl.ArrayMaker(; regions, values, args) => begin
-            isempty(args) || return ROArgsT{T}(args)
-            push!(args, BSImpl.Const{T}(regions))
-            push!(args, BSImpl.Const{T}(values))
-            return ROArgsT{T}(args)
+            cached = _get_cached_arguments(args)
+            cached === nothing || return ROArgsT{T}(cached)
+
+            newargs = ArgsT{T}()
+            push!(newargs, BSImpl.Const{T}(regions))
+            push!(newargs, BSImpl.Const{T}(values))
+            return ROArgsT{T}(_publish_cached_arguments!(args, newargs))
         end
     end
 end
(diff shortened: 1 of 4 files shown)


AayushSabharwal:
Wow, nice catch. While the fix looks correct, I think this might cause problems downstream for users that access `args` directly. Not sure if we still have those.

AayushSabharwal:
Well if it's all green then I'm happy to be wrong

AayushSabharwal (approved):

oameye:
Thanks for the fast merge ❤️

oameye:
Can we release this?

oameye:
@AayushSabharwal @ChrisRackauckas Hey kinda need this for QC.jl. Can you tag a release?

oameye:
Thanks Chris!

PR #1068, opened 2026-09-09
fix(polyform): support array-form MSE expressions

ajatshatru01:
Treat ArrayOp and ArrayMaker nodes as opaque polynomial variables during polynomial conversion, allowing scalar coefficients in symbolic array expressions to be expanded safely.

Add a regression test for expanding a weighted MSE loss constructed from a symbolic array.

### DIFF
--- a/src/polyform.jl
+++ b/src/polyform.jl
@@ -20,6 +20,14 @@ function basicsymbolic_to_polyvar(bs_to_poly::AbstractDict, x::BasicSymbolic)::P
     end
 end
 
+function opaque_symbolic_to_polyvar(
+        bs_to_poly::AbstractDict, x::BasicSymbolic, inner_name::Symbol)::PolyVarT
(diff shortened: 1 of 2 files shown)


ajatshatru01:
since the parser i am working on requires array form mse cost and not scalarized cost, i am making this change

ajatshatru01:
#https://github.com/SciML/NeuralPDE.jl/pull/1140

src/polyform.jl
```diff
@@ -20,6 +20,14 @@ function basicsymbolic_to_polyvar(bs_to_poly::AbstractDict, x::BasicSymbolic)::P
     end
 end
 
+function opaque_symbolic_to_polyvar(
```
AayushSabharwal, line 23: This is unnecessary. `basicsymbolic_to_polyvar` exists.

AayushSabharwal (changes requested):

PR #1069, opened 2026-09-10
Widen polyform gcd coefficients to Int64 (32-bit)

ChrisRackauckas-Claude:
## Summary
- On 32-bit Julia, homogeneous `Int32` coeffs survive `poly_to_gcd_form` via `Integer.(...)` and then overflow in MultivariatePolynomials `gcd`/`div_multiple` (`DivideError`).
- Always widen integer/rational coeffs to `Int64` / `Rational{Int64}` before gcd.
- Replaces closed #1065; bumps to **4.46.5** for registration.
- Unblocks MomentClosure x86 (bernoulli / derivative matching).

## Test plan
- [ ] `test/polyform.jl` including homogeneous Int32 case
- [ ] MomentClosure x86 Core green against this tip

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SymbolicUtils"
 uuid = "d1185830-fcd6-423d-90d6-eec64667417b"
-version = "4.46.4"
+version = "4.46.5"
 authors = ["Shashi Gowda"]
 
 [deps]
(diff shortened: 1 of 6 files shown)


ChrisRackauckas-Claude:
Pushed another Aqua fix: include `UpperHessenberg` in `_StructuredMatrix` under `@static if VERSION >= v\"1.13\"` instead of a standalone method (the standalone form left 4 ambiguities against LinearAlgebra on 1.13).

PR #1077, opened 2026-09-13
fix: restrict nested power folding (x^a)^b to integer b to prevent unsound simplifications

### LINKED ISSUE #1045: `(x^2)^(1//2)` folds to `x`, dropping the absolute value for negative real `x`
Hello,

`(x^2)^(1//2)` is folded to `x`, which is wrong for negative real `x`. The
symbolic result then disagrees with evaluating the same expression in plain
Julia, and the disagreement survives into generated code.

## MWE

SymbolicUtils v4.46.2 (current release), Julia 1.12.7, on a symbol declared
`Real` so that no branch-cut convention is in play:

```julia
julia> using SymbolicUtils

julia> @syms r::Real
(r,)

julia> symtype(r)
Real

julia> (r^2)^(1//2)
r

julia> isequal((r^2)^(1//2), r)
true
```

For `r = -3` the correct value is `3`, not `-3`. Through Symbolics
v7.39.0, where the consequence is visible:

```julia
julia> using Symbolics

julia> @variables x
1-element Vector{Num}:
 x

julia> substitute((x^2)^(1//2), Dict(x => -3.0))
-3.0                       # ← wrong

julia> ((-3.0)^2)^(1//2)   # the same expression, plain Julia
3.0
```

It reaches generated code too:

```julia
julia> f = eval(build_function((x^2)^(1//2), [x]; expression = Val{true}));

julia> f([-3.0])
-3.0
```

## Why this is not a display or convention question

`(x^a)^b = x^(a*b)` holds for real `x` only when `x ≥ 0` (or when `b` is an
integer). The folding is applied unconditionally, so it silently drops the
absolute value. The rule is correct and welcome when the outer exponent is
an integer:

```julia
julia> (x^2)^2
x^4                        # fine
```

The same defect with a different exponent turns a well-defined real value
into a domain error:

```julia
julia> (x^2)^(1//4)
x^(1//2)                   # at x = -3: sqrt(-3), while ((-3)^2)^(1/4) = 1.7320508…
```

And the same mathematical object written the other way is left alone, so the
two spellings disagree:

```julia
julia> sqrt(x^2)
sqrt(x^2)                  # untouched
```

`(x^2)^0.5` folds to `x` as well.

## Suggested fix

Apply `(x^a)^b → x^(a*b)` only when `b` is an integer, or when `x` is known
non-negative. Without a way to state the latter, the integer condition is
the only sound one; `sqrt(x^2)` shows the conservative behaviour is already
implemented elsewhere.

This is also, concretely, what the absence of an assumptions framework
costs (JuliaSymbolics/Symbolics.jl#98): the gap does not merely block
simplifications that would need `x > 0`, it lets an unsound one through by
default. `abs(x)` would be the correct result here and is what a CAS with
assumptions returns.

## Context

Found by differentially testing simplification of ~30 elementary identities
against another CAS, while auditing the uncertainty expressions of a
symbolic GUM library, where a dropped absolute value is a wrong measurement
result rather than a cosmetic difference.

*Disclosure: this report was written with AI assistance, and the reduction
to this MWE was AI-assisted as well. Every snippet above was executed on the
stated versions and its output pasted from the actual session.*

s-celles:
Currently, the `pow` constructor unconditionally folds `(x^a)^b` into `x^(a*b)`. While this is sound for integer `b`, it drops absolute values for fractional powers (e.g. `(x^2)^(1/2)` becomes `x` instead of `abs(x)`). This breaks downstream packages relying on strict math properties (like GUM uncertainty propagation). 
This PR adds `safe_isinteger(b)` to the folding rule so that fractional outer powers are preserved un-evaluated, letting higher-level simplification rules handle them properly.

Fixes #1045

Assisted by: AI


AayushSabharwal (commented):
Thanks! Please add a quick test so we can ensure this doesn't regress.

s-celles:
Currently quite busy this end of week... but if you are fine with the "big picture" exposed https://github.com/JuliaSymbolics/SymbolicUtils.jl/issues/1045#issuecomment-5653040078 about assumptions please take the lead on this.

s-celles:
I've added the quick tests for fractional and nested powers to `test/rulesets.jl` as requested to ensure this doesn't regress.

ChrisRackauckas:
test fails

test/basics.jl
```diff
@@ -956,7 +956,7 @@ end
     @syms a b c
     t = SymbolicUtils.maketerm(typeof(b + c), +, [a,  (b+c)], nothing)
     @test isequal(t.dict, ACDict{SymReal}(a => 1, b => 1, c => 1))
-    @test isequal(SymbolicUtils.maketerm(typeof(b^2), ^, [b^2,  1//2],  nothing), b)
```
s-celles, line 959: testing `(b^2)^(1//2) = b` is mathematically wrong when b < 0 (should be `abs(b)`)

so we test integer exponent folding instead

s-celles: tests on main were also failing so I'm not really sure if all is good.

AayushSabharwal (approved):

s-celles:
Thanks @AayushSabharwal for merging

PR #1078, opened 2026-09-13
Support complex symbolic numeric domains

oameye:
This is the small SymbolicUtils prerequisite for the companion Symbolics.jl complex-number redesign.

The companion PR introduces the distinction

```julia
Num <: Real
SymbolicNumber <: Number
```

so that a symbolic scalar which may be complex can remain an ordinary symbolic expression instead of being forced into `Complex{Num}`. For that to work, SymbolicUtils must preserve the distinction between expressions known to be `Real` and expressions only known to be `Number`.

This PR provides that infrastructure. The changes are generic rather than complex-specific: numeric domains are preserved through arithmetic and simplification, polynomialization accepts general numeric coefficients, and `poly_to_gcd_form` preserves concrete exact coefficient domains. The latter also fixes an independently exposed regression where `Rational{BigInt}` could be narrowed to `Rational{Int64}` in exact polynomial/GCD paths, while retaining the existing safe widening of narrow integer types where required by the polynomial backend.

Production diff:

```text
src/:   +105 / -14
tests/: +113 / -1
```

There is no parallel complex algebra or complex polynomial implementation.

The substantive motivation, representation change, examples, breaking implications, historical issue coverage, and discussion belong in the companion [Symbolics PR](https://github.com/JuliaSymbolics/Symbolics.jl/pull/1985):


src/polyform.jl
```diff
@@ -106,7 +106,25 @@ function to_poly!(poly_to_bs::AbstractDict, bs_to_poly::AbstractDict, expr::Basi
...
+                # `re + im * im_part`; lowering it here lets equivalent forms cancel
+                # without relaxing the constructor's real-component contract.
+                rpoly = to_poly!(poly_to_bs, bs_to_poly, args[1], recurse)
+                ipoly = to_poly!(poly_to_bs, bs_to_poly, args[2], recurse)
+                poly = zeropoly()
+                MA.operate!(+, poly, rpoly)
+                # Promote either a polynomial variable or a polynomial into the
+                # package's wide `PolynomialT` coefficient storage before multiplying by
+                # the literal Julia `im`. This avoids specializing the polynomial on a
+                # concrete Complex coefficient type.
```
ChrisRackauckas: Are we sure this is the better way to handle it?

oameye: Good point. This came from #921/#922.

The original issue was that `Term(complex, [re, im])` is the canonical explicit-Cartesian representation used by Symbolics for `Complex{Num}`. We did not want to rewrite that globally to `re + im*im`, because direct `real`/`imag`/`conj` extraction and round-tripping rely on that node remaining intact.

For example, at the symbolic level we want

```julia
c = complex(x, y)

real(c) == x
imag(c) == y
```

without first expanding `c` into an ordinary sum/product tree.

At the same time, treating `complex(re, im)` as an opaque `PolyVar` inside `to_poly!` is algebraically wrong for canonicalization. For example,

```julia
complex(0, x)
im*x
```

are the same scalar, so

```julia
simplify(complex(0, x) - im*x; expand = true)
```

should reduce to `0`.

Likewise,

```julia
complex(x, y) - (x + im*y)
```

should vanish once we explicitly ask for polynomial expansion.

The earlier #922 fix therefore interpreted the Cartesian node algebraically only at the `to_poly!` boundary, rather than changing its symbolic representation globally.

Looking at it again in the context of this redesign, I still think that boundary is the right one. General complex symbolic scalars are now atomic `Number`-valued expressions and no longer use `Term(complex, ...)`; this case is only for the explicit Cartesian representation.

So the intended behavior is:

```text
symbolic IR:
    complex(x, y)          stays explicit Cartesian

polynomial IR:
    complex(x, y)          lowers to x + im*y
```

This lets us have both

```julia
real(complex(x, y)) == x
```

and

```julia
simplify(complex(x, y) - (x + im*y); expand = true) == 0
```

without globally decomposing complex symbolic expressions.

I do agree that the current implementation makes that simple rule look more complicated than it should. I’m going to tighten it to the ordinary scalar two-real-component constructor and simplify the code/comments.

PR #1080, opened 2026-09-13
Make calling a symbolic function non-differentiable

utkuyilmaz1903:
Calling a symbolic function, `u(t, x)` for `@syms u(::Real, ::Real)`, returns a term: the arguments are stored in it, not computed with. Under Zygote the call was traced into term construction and failed, so the symbol had to be built outside the differentiated function, e.g. `sol[u(t, x)]` in a loss on a solution (SciML/MethodOfLines.jl#309). With the rule the call is opaque to reverse-mode AD, so it no longer has to be hoisted.


ChrisRackauckas:
I'm not sure if this is true? `u(p,x)` what happens in this case?

utkuyilmaz1903:
`u(0.3, x)` is still a term, not a number: the arguments are stored in it, not computed with, so the call itself has nothing to differentiate. Turning a term back into a number is `substitute`/`build_function`, which this rule doesn't touch.

ChrisRackauckas:
https://github.com/JuliaSymbolics/SymbolicUtils.jl/pull/1086 is a safer form.

Issue #1079, opened 2026-09-13
simplify replaces exact sqrt terms with floating point numbers


s-celles:
`simplify` folds an unevaluated `sqrt` term over an exact argument into a floating point number, which loses exactness that the input had.

## Reproducer

```julia
julia> using SymbolicUtils

julia> s = SymbolicUtils.term(sqrt, 2)
sqrt(2)

julia> simplify(2s)
2.8284271247461903     # expected 2sqrt(2)

julia> simplify(s^2)
2.0                    # expected 2

julia> simplify(s*s)
2.0                    # expected 2

julia> simplify(s^4)
4.0                    # expected 4
```

The folding is not applied consistently, which is what makes it look unintended rather than a deliberate constant-folding policy:

```julia
julia> simplify(s^3)
sqrt(2)^3              # left exact

julia> simplify(s - s)
0                      # exact

julia> simplify(s / s)
1                      # exact

julia> simplify((s * SymbolicUtils.term(sqrt, 3))^2)
3.0*2.0                # folded to floats, and not combined either
```

So `s^2` and `s^4` become floats while `s^3` stays symbolic, and `2s` becomes a float while `s - s` and `s/s` stay exact.

## Why it matters

`term(sqrt, n)` is the natural way to carry an exact algebraic constant like `sqrt(2)` through an expression when the value has no exact `Float`-free representation otherwise. Folding it to a `Float64` means any downstream `simplify` silently destroys the exactness.

We hit this in [SymbolicIntegration.jl](https://github.com/JuliaSymbolics/SymbolicIntegration.jl): antiderivatives of rational functions needing an algebraic extension are exact, for example

```
(1//4)*log(sqrt(2) - x)*sqrt(2) - (1//4)*log(sqrt(2) + x)*sqrt(2)
```

but they cannot be verified by simplifying `d/dx F - f` to zero, because `simplify` evaluates `sqrt(2)` in some subterms and not in others, leaving a nonzero residual for an expression that is in fact identically zero.

## Expected

Either leave `sqrt` over an exact argument unevaluated (so `simplify(s^2) == 2` and `simplify(2s) == 2sqrt(2)`), or at least apply the same rule in all of these cases so the behaviour is predictable.

## Versions

Julia 1.13.0, SymbolicUtils 4.46.6. Reproduced with SymbolicUtils alone, no Symbolics needed.

---

*Disclosure: this issue was investigated and written with AI assistance. The reproducer was run and verified before filing.*

PR #1081, opened 2026-09-13
Add _extract_perfect_square for building an exact square root

s-celles:
Relates to #1079 and required by https://github.com/JuliaSymbolics/SymbolicIntegration.jl/pull/136.

## Problem

There is currently no way to put an exact square root into a symbolic expression. `Base.sqrt` returns a floating point number for an integer or rational argument, so the approximation is baked in before any symbolic machinery sees it:

```julia
julia> sqrt(3)
1.7320508075688772
```

Anything derived from that carries the rounding. `term(sqrt, 3)` does build an unevaluated term, but it leaves the radicand unreduced and it is not an obvious spelling for "the exact square root of 3".

## What this adds

```julia
julia> exact_sqrt(2)
sqrt(2)

julia> exact_sqrt(12)
2sqrt(3)

julia> exact_sqrt(12 // 49)
(2//7)*sqrt(3)

julia> exact_sqrt(7 // 2)
(1//2)*sqrt(14)

julia> exact_sqrt(4)
2
```

A perfect square gives back an exact number rather than a symbolic expression, a rational radicand is rationalised, and a negative argument throws `DomainError` to match `Base.sqrt` on a real number.

The perfect square factor is found by trial division bounded by `PERFECT_SQUARE_TRIAL_LIMIT`, followed by a perfect-square test on the remaining cofactor, so a large radicand cannot make the search run away. Any factorisation with `n == s^2*r` is a correct answer and a larger `s` only yields a tidier radical, so bounding the search costs exactness nothing — it only means a radicand built from two large primes stays unreduced.

## Why here

This came out of [SymbolicIntegration.jl](https://github.com/JuliaSymbolics/SymbolicIntegration.jl), where antiderivatives of rational functions needing an algebraic extension were coming back as floats — `0.5773502691896257` instead of `(1//3)*sqrt(3)` (JuliaSymbolics/SymbolicIntegration.jl#135). The fix there needed exactly this primitive, and it is currently carried as a private helper in that package. It is not integration-specific, so it seems better placed here, where anyone building an exact symbolic constant can reach it.

## Scope

This is the constructor only. It does not address the folding reported in #1079 — `simplify(exact_sqrt(2)^2)` still gives `2.0` rather than `2`, and `simplify(2*exact_sqrt(2))` still gives `2.8284271247461903`. Those are a separate matter; this PR just makes it possible to *build* the exact value in the first place. Happy to rename (`symbolic_sqrt`, `sqrt_exact`, …) or to move it out of the exported API if you would rather keep it internal.

## Tests

Added to `test/methods.jl`: perfect squares returning exact numbers, irrational cases returning unevaluated terms, the perfect square factor being pulled out for integer and rational radicands, a numeric check that each result equals the square root it claims to be, a radicand whose prime factors all lie beyond the trial division bound, and the `DomainError` cases. `test/methods.jl` passes in full (the new testset is 27 tests). The docstring examples are `jldoctest` and match actual output.

---

*Disclosure: this change was written with AI assistance, and tested locally before opening the PR. `test/methods.jl` was run in full, and every figure quoted above is copied from an actual run.*


docs/src/api.md
```diff
@@ -6,6 +6,7 @@
 ```@docs; canonical=false
 SymbolicUtils.@syms
 SymbolicUtils.term
+SymbolicUtils.exact_sqrt
```
ChrisRackauckas, line 9: Seems a bit weird to put it on this level. We should probably put it as some mathematical operations set.

src/methods.jl
```diff
@@ -2104,3 +2104,90 @@ end
...
     return Array{safe_eltype(Tarr), length(Tidxs)}
 end
+
+"""
+    exact_sqrt(x::Union{Integer, Rational})
+
+Square root of the nonnegative `x`, as an exact symbolic expression.
+
+`Base.sqrt` returns a floating point number for an integer or rational
+argument, so an expression built from it can no longer be exact:
```
AayushSabharwal: This is because `sqrt(3)` doesn't hit any SymbolicUtils dispatch. It doesn't build an expression because there is no symbolic computation in that function-call. Simple `sqrt(SymbolicUtils.Const{SymReal}(3))` (or the shorthand `sqrt(Symbolics.SConst(3))`) will avoid this.

src/methods.jl
```diff
@@ -2104,3 +2104,90 @@ end
...
+    x >= 0 || throw(DomainError(x, "exact_sqrt requires a nonnegative argument"))
+    iszero(x) && return 0
+    # sqrt(num//den) == sqrt(num*den)//den
+    num, den = big(numerator(x)), big(denominator(x))
+    s, r = _split_perfect_square(num * den)
+    c = _narrow(s // den)
+    isone(r) && return c
+    radical = term(sqrt, _narrow(r))
+    isone(c) ? radical : c * radical
+end
```
AayushSabharwal: This function is type-unstable. Depending on the value of `x`, it could return an `Int` (the literal `0`) or a rational or an expression.

src/methods.jl
```diff
@@ -2104,3 +2104,90 @@ end
...
+```
+
+A perfect square gives back an exact number rather than a symbolic expression.
+Negative arguments throw, matching `Base.sqrt` on a real number.
+"""
+function exact_sqrt(x::Union{Integer, Rational})
+    x >= 0 || throw(DomainError(x, "exact_sqrt requires a nonnegative argument"))
+    iszero(x) && return 0
+    # sqrt(num//den) == sqrt(num*den)//den
+    num, den = big(numerator(x)), big(denominator(x))
```
AayushSabharwal: Why eagerly bigify?

src/methods.jl
```diff
@@ -2104,3 +2104,90 @@ end
...
+Prime factors are removed by trial division up to
+[`PERFECT_SQUARE_TRIAL_LIMIT`](@ref), then the remaining cofactor is tested for
+being a perfect square itself. Any pair with `n == s^2*r` is a correct answer; a
+larger `s` only yields a tidier radical, so bounding the search costs exactness
+nothing.
+"""
+function _split_perfect_square(n::Integer)
+    s, r = big(1), big(n)
+    p = big(2)
+    while p <= PERFECT_SQUARE_TRIAL_LIMIT && p * p <= r
```
AayushSabharwal: You only need to iterate over primes, so this could be improved by just having a global list of primes up to a limit.

src/methods.jl
```diff
@@ -2104,3 +2104,90 @@ end
...
+        end
+        p += 1
+    end
+    q = isqrt(r)
+    if q * q == r
+        s *= q
+        r = big(1)
+    end
+    return (s, r)
+end
```
AayushSabharwal, line 2174: Why not just
```julia
function _split_perfect_square(n::Integer)
  n = big(n)
  s = floor(BigInt, sqrt(n))
  r = n - s*s
end
```

AayushSabharwal (changes requested):
Apart from these comments, I'm not entirely convinced this `exact_sqrt` function is necessary. You can avoid evaluating `sqrt(3)` by wrapping it as suggested in a comment. That enables retaining the exact representation. `exact_sqrt` doesn't need to fold out square factors, that can be done as a postprocessing pass on the entire expression. Sure, this might make it print slightly nicer, but does it offer any other benefit?

src/methods.jl
s-celles: `n = s^2 + r` should be `n = s^2 × r`

AayushSabharwal: Right my bad

AayushSabharwal (approved):

s-celles:
Thanks @ChrisRackauckas for merging

PR #1084, opened 2026-09-15
Emit static arrays for every small array in ArrayMaker/array_literal codegen

### LINKED ISSUE #1083: Small arrays only become static arrays when every dimension is at most 4 and rank is at most 2
`ArrayMaker`/`array_literal` codegen scalarizes any array with at most `FILL_ARR_LIMIT` (16) elements. With the default allocator it can emit a `StaticArray` through `fill_sarr` rather than allocating a buffer and filling it with `fill_arr!`, which avoids the allocation.

The guard for that is `all(<=(STATIC_ARRAY_LIMIT), sz) && ndims <= 2` with `STATIC_ARRAY_LIMIT = 4`. That bounds each dimension rather than the element count, so most small shapes still allocate:

| shape | elements | emitted |
| --- | --- | --- |
| `(3,)` | 3 | `fill_sarr` |
| `(3, 3)` | 9 | `fill_sarr` |
| `(4, 4)` | 16 | `fill_sarr` |
| `(8,)` | 8 | `zeros` + `fill_arr!` |
| `(5, 3)` | 15 | `zeros` + `fill_arr!` |
| `(16,)` | 16 | `zeros` + `fill_arr!` |
| `(2, 2, 2)` | 8 | `zeros` + `fill_arr!` |

## Reproducer

```julia
using SymbolicUtils, SymbolicUtils.Code
using SymbolicUtils: Const, SymReal

@syms a::Real b::Real

w = @makearray w[1:8] begin
    w[1:8] => Const{SymReal}([a, b, a + b, a - b, 2a, 2b, a * b, a / b])
end

Code.fast_toexpr(w, Dict{Any, Any}())
```

```
begin
    __miscₛᵧₘ0 = zeros
    local var"##cse#0" = __miscₛᵧₘ0((8,))
    ...
    __miscₛᵧₘ1 = (SymbolicUtils.Code.fill_arr!)(var"##cse#0", (Val)((8,)), ...)
end
```

Two things stand in the way of widening this:

- `STATIC_ARRAY_LIMIT`'s docstring says "Maximum total element count", so the per-dimension check does not match what it documents. Applying it as a total would instead shrink the current behaviour, since the `(4, 4)` case above has 16 elements.
- `fill_sarr` only handles 1D and 2D and errors on anything else, which is why rank 3 is excluded.

Everything reaching this branch is already capped at 16 elements, and `SArray{Tuple{sz...}}` covers any rank, so the element-count check should be redundant with `FILL_ARR_LIMIT`.

Downstream context: JuliaComputing/DyadCompilerPasses.jl#81, where a `fill_arr!` that stays a regular array causes type instability that `HVNCAT_STATIC_RULE` would otherwise remove.

AJ0070:
Closes #1083

`fill_sarr` only kicked in for rank <= 2 with every dimension <= 4, so `(8,)`, `(16,)`, `(5, 3)` and anything 3D still allocated a buffer and filled it with `fill_arr!`, even though the branch they land in already caps the array at 16 elements.

`fill_sarr` now builds `SArray{Tuple{sz...}}` instead of special-casing `SVector`/`SMatrix`, so it handles any rank (1D and 2D still come out as `SVector`/`SMatrix`, they're aliases). `STATIC_ARRAY_LIMIT` now bounds the total element count, which is what its docstring always described, rather than each dimension separately. It stays at 16 so that nothing static today stops being static: the `(4, 4)` case the old check allowed is already 16 elements.

`with_allocator` and `Returns(buf)` still go through `fill_arr!`, since the result has to land in the caller's buffer.

The tradeoff is that generated functions now hand back an `SArray` for anything up to 16 elements instead of up to 4 per dimension. That's the same call 709cfaa made, just applied to the rest of the shapes.

Tests in `test/new_code.jl` cover `fill_sarr` at rank 1, 2 and 3, and check that the shapes that changed now codegen to it and still agree with `Code.toexpr`. Core suite is green locally: 18134 passed, 3 broken, all three pre-existing in the `ArrayOp` testset.

Patch bump to 4.46.7.


src/faster_code.jl
```diff
@@ -582,7 +571,7 @@ function codegen_function!(::Type{ArrayMaker{T}}, cs::CodegenState{T}, expr::Bas
         for ax in sh
             push!(sz_expr.args, length(ax))
         end
-        if all(<=(STATIC_ARRAY_LIMIT), sz_expr.args) && length(sz_expr.args) <= 2 && _allocator === zeros
+        if _allocator === zeros
```
AayushSabharwal: This should still be gated behind a length check to not create massive SArrays

src/faster_code.jl
```diff
@@ -763,8 +752,7 @@ function codegen_function!(
 
     if len <= FILL_ARR_LIMIT
         sz_val = unwrap_const(cs.ir[first(args_idxs)])
-        ndims_arr = length(sz_val)
-        if all(<=(STATIC_ARRAY_LIMIT), sz_val) && ndims_arr <= 2 && _allocator === zeros
+        if _allocator === zeros
```
AayushSabharwal: Same here

AayushSabharwal (changes requested):

src/faster_code.jl
AJ0070: Added `STATIC_ARRAY_LIMIT` back for this, so the static path doesn't just inherit whatever `FILL_ARR_LIMIT` happens to be. It now caps the total element count, which is what its docstring always described, instead of each dimension separately.

Left it at 16 because that's what master already emits: the `(4, 4)` case the old per-dimension check allowed is 16 elements, so anything lower would turn shapes that are static today back into heap arrays. Happy to bring it down if you'd rather have a tighter ceiling, it just costs some of what's already there.

src/faster_code.jl
AJ0070: Same check added here.

AayushSabharwal (approved):

AJ0070:
@AayushSabharwal Can you please re-trigger CI?

DhairyaLGandhi:
Bump @AayushSabharwal :)

PR #1085, opened 2026-09-16
Substitute ArrayOp indices inside Operator terms during scalarization

oscardssmith:
reduce_eliminated_idxs used the default substitute filter, which refuses to recurse into Operator terms. scalarize(sum(D(x))) thus left the placeholder index in place and collapsed to 3D(x[_1]). Share the ArrayOp scope filter already used for output indices.

Fixes JuliaSymbolics/Symbolics.jl#1981.

Coauthored by fable.
cc @kpamnany


kpamnany (approved):
Confirming that this fixes https://github.com/JuliaSymbolics/Symbolics.jl/issues/1981 and also the STK issue that caused that report. Thanks!

